PR #56, opened 2015-08-22
Translate @erikerlandson's NonNegInt example to refined

fthomas:


fthomas:
The rendered markdown can be seen here: https://github.com/fthomas/refined/blob/topic/nonnegint/docs/nonnegint.md

Issue #14, opened 2015-06-06
Replace String with RefinementError as error type


fthomas:
`refine` currently returns an `Either[String, A]`. `String` is an awful type for errors and should probably be replaced by a dedicated error type like `RefinementError`.

```scala
case class RefinementError(msg: String)
```

fthomas:
Even `RefinementError` is not good enough. Suppose we have a predicate `A And B`. How do I know if only `A` failed, only `B` failed or both failed? (Parsing the error message is of course not an option.)

fthomas:
I started working on this here: https://github.com/fthomas/refined/compare/wip/errors-as-data

fthomas:
Current WIP branch is https://github.com/fthomas/refined/compare/wip/errors-as-data-4

fthomas:
The -4 branch has been superseded by https://github.com/fthomas/refined/compare/wip/errors-as-data-5 which is pretty close to being merged to master. I'll open a PR once the last rough edges are gone.

Anyhow, the branch is stable enough to play around with the new representation of validation results. You could start playing by evaluating
```scala
Validate[Unit, True Or False].validate(())
```
which should return `Passed(Or(Passed(True()), Failed(False())))` with type `Result[Or[Result[True], Result[False]]]`.

Issue #26, opened 2015-07-12
Lift the "only literals" constraint when the predicate is constant


fthomas:
```scala
scala> val x: List[Int] @@ True = List(1, 2, 3)
<console>:40: error: refineLit only supports literals
       val x: List[Int] @@ True = List(1, 2, 3)
                                      ^
```
The rhs is not required for evaluation of the predicate (because it is constant), so this refinement should be possible at compile-time. The same is true for `val x: List[Int] @@ Not[False] = List(1, 2, 3)`.

fthomas:
wow, much branch, very wip:

* https://github.com/fthomas/refined/tree/wip/isConstant
* https://github.com/fthomas/refined/tree/wip/predicate0
* https://github.com/fthomas/refined/tree/wip/ConstPredicate

fthomas:
Fixed by https://github.com/fthomas/refined/commit/a1859eeedc798de3111ad53fc0b719179e4ac93c.

PR #62, opened 2015-10-03
Support a richer encoding of validation results

### LINKED ISSUE #61: Consider renaming implicits object to auto
The implicits in `implicits` are intended for automatic conversions from base types to refined types or for automatic conversions between refined types. There are also more implicits in the library which are not in `implicits` and that are not used for automatic conversions (e.g. all the type class instances).

Observe also that all functions in `implicits` already begin with `auto`.

fthomas:
This commit is an important step towards fixing #14:

- Change all predicates so they can be represented as data. That means
  `trait`s are now `case class`es that can hold values of their type
  parameters.

- Add a `Result[A]` ADT (with cases `Passed` and `Failed`) that
  represents the result of a validation against a type-level predicate.
  A `Result` contains some detail about the validation of type `A`
  which allows arbitrary nested results.

- Replace the `Predicate` type class with `api.Validate`. `Validate`'s
  main function is `validate` which has type `T => Result[R]` where `T`
  is the type of the value that is validated and `R` (which is an
  abstract type member of `Validate`) represents the detail of the
  validation result. For simple (non-nested) predicates `R` equals the
  type of the predicate `P`.

It also contains some cleanup:

- Move `InferenceRule`, `Refined`, and `RefTyp` into the new `api`
  package.

- Rename `InferenceRule` to `Inference`.

- Add a `internal.Resources` object that contains all user-visible
  strings.

@koshelev if you're still interested in this, I'd very much appreciate if you could review this.


koshelev:
@fthomas I'm definitely still interested in it and will be happy to review PR. Sorry, was buried into projects at work for several past weeks

fthomas:
Cool, thanks! The most important parts are `Validate`, `Result`, and the implementation of the `Validate` instances. Note that I intentionally didn't change the signature of `refineT` and `refineV` yet (they still return an `Either[String, X]`). My focus was on how to change `Predicate` to support validation results as values.

I'm also open to suggestions for alternative / better names for `Result.detail`, `Validate.R`, and `Validate.Res`. They could be more expressive.

shared/src/main/scala/eu/timepit/refined/api/Validate.scala
```diff
@@ -0,0 +1,113 @@
...
+trait Validate[T, P] extends Serializable { self =>
+
+  type R
+
+  final type Res = Result[R]
+
+  def validate(t: T): Res
+
+  /** Returns a string representation of this [[Validate]] using `t`. */
+  def showExpr(t: T): String
```
koshelev: I would prefer to have a type class (`Show` from `cats` for instance) instead of `showExpr` and `showResult`

shared/src/main/scala/eu/timepit/refined/api/Validate.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  def showResult(t: T, r: Res): String =
+    Resources.predicateResultDetailDot(r, showExpr(t))
+
+  /**
+   * Denotes whether this [[Validate]] is constant (which is false by
+   * default). A constant [[Validate]] ignores the argument passed to
+   * [[validate]].
+   */
+  val isConstant: Boolean = false
```
koshelev: Why is it important to know, is a validation constant or not?

fthomas: The reason is outlined in #26. See also https://github.com/fthomas/refined/blob/master/docs/ctorfieldnames.md

I admit that this might not look very compelling but it is kind of cool that the `ConstructorNames` and `FieldNames` predicates can be evaluated at compile-time. :) But this is also present in the current `Predicate` and nothing new that has been introduced in this PR.

shared/src/main/scala/eu/timepit/refined/api/Validate.scala
fthomas: I tried this separation in the other `wip/errors-as-data-*` branches but was overwhelmed by the amount of changes this entails. I still agree that the separation is useful but would like to defer it to a later PR.

shared/src/main/scala/eu/timepit/refined/api/Validate.scala
koshelev: Seems like I'm missing something... What is relation between `isConstant` field and compile time predicates?

shared/src/main/scala/eu/timepit/refined/api/Validate.scala
koshelev: makes sense

shared/src/main/scala/eu/timepit/refined/api/Validate.scala
fthomas: https://github.com/fthomas/refined/blob/master/shared/src/main/scala/eu/timepit/refined/internal/RefineMAux.scala#L29

If `isConstant` is true we can call `validate` without requiring a literal value.

koshelev: macros, got it

koshelev:
LGTM

shared/src/main/scala/eu/timepit/refined/api/Validate.scala
koshelev: Still, is it possible to have it on type level?

fthomas: You mean adding a designated `ConstValidate` type for predicates that can be evaluated without an input? Something like this would be nice and I tried to do it. The problem are higher-order predicates like `Not[P]`. If `P` is constant `Not[P]` should be constant too, but if it isn't `Not[P]` should also not be constant. Adding the `isValid` field was the only viable solution I could come up with. But I'm open for other suggestions. :-)

I could also be persuaded to remove the notion of constant predicates completely because the examples we have are not very compelling.

fthomas:
@koshelev Many thanks for the review!

Issue #58, opened 2015-08-28
Add ScalaCheck support


fthomas:
From the Gitter channel

> @koshelev Another thing is to add some scalacheck support, but it should not be so hard
> @fthomas The ScalaCheck support sounds interesting. I guess you want to have an `Arbitrary[F[T, P]: RefType]` automatically if there is an `Arbitrary[T]` and a `Predicate[P, T]`?

Filtering an existing `Arbitrary[T]` with a `Predicate[P, T]` will discard too many values in most cases, so that is probably too simplistic. We should nevertheless check how to make working with refined and ScalaCheck nicer.

fthomas:
https://github.com/julien-truffaut/Monocle/commit/7415e0791e30c7de3333c9810d4d56601158aa3b has a compelling use case. It contains
```scala
type Minute = Int Refined Interval[W.`0`.T, W.`59`.T]

implicit val minuteArbitrary: Arbitrary[Minute] = Arbitrary(
  Gen.choose(0, 59).map(Refined.unsafeApply)
)
```

In this case refined should be able to provide a `Arbitrary[Minute]` automatically. This would also prevent the repetition of `0` and `59` (which are details of the `Minute` predicate) in the tests.

fthomas:
#73 has been merged, so the next release will include a refined-scalacheck module with `Arbitrary` instances for some of our predicates.

PR #73, opened 2015-10-14
First draft of a scalacheck module

fthomas:
This is just a draft of a possible `refined-scalacheck` module, see #58. It only has an `Arbitrary` instance for `F[T, Interval[H, L]]` but there are a few others which should be just as easy to write (e.g. `Letter`, `Digit`, `EndsWith`, `StartsWith`, `Empty`, `Size`, etc.). 

@koshelev since you was the first to ask for this, do you have any specific predicates in mind for which we should provide `Arbitrary` instances?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,10 @@
 lazy val root = project.in(file("."))
-  .aggregate(coreJVM, coreJS, docs)
+  .aggregate(
+    coreJVM,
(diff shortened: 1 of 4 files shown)


fthomas:
I'm pretty satisfied how this module looks right now and am going ahead and merge it.

PR #77, opened 2015-10-21
Use macro-compat to enable -Xfatal-warnings again

fthomas:


fthomas:
This seems not as simple as I thought. Closing for now.

PR #82, opened 2015-10-28
Add Eval predicate

fthomas:
This adds the `Eval` predicate that takes a `String` singleton type as parameter. The value of that singleton type is a definition of a predicate `T => Boolean` which is parsed and evaluated at compile-time to check if a given value satisfies it. `Eval` allows to define arbitrary predicates without writing `Validate` instances.

Example usage:
```scala
scala> type IsEven = Eval[W.`"(x: Int) => x % 2 == 0"`.T]
defined type alias IsEven

scala> val a: Int @@ IsEven = 4
a: shapeless.tag.@@[Int,IsEven] = 4

scala> val b: Int @@ IsEven = 5
<console>:47: error: Predicate failed: (x: Int) => x % 2 == 0.
       val b: Int @@ IsEven = 5
                              ^
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,7 @@ The library comes with these predefined predicates:
 [`generic`](https://github.com/fthomas/refined/blob/master/core/shared/src/main/scala/eu/timepit/refined/generic.scala)
 
 * `Equal[U]`: checks if a value is equal to `U`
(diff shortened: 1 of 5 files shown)


fthomas:
Apparently this is JVM only.

PR #106, opened 2015-12-14
[WIP] Revamp instances of Arbitrary for Numeric

jedesah:
- Remove hardcoded Int
- Add instances for shapeless.Nat

### DIFF
--- a/contrib/scalacheck/shared/src/test/scala/eu/timepit/refined/scalacheck/NumericArbitrarySpec.scala
+++ b/contrib/scalacheck/shared/src/test/scala/eu/timepit/refined/scalacheck/NumericArbitrarySpec.scala
@@ -1,13 +1,16 @@
 package eu.timepit.refined
 package scalacheck
 
-import eu.timepit.refined.api.Refined
+import eu.timepit.refined.api.{ RefType, Validate, Refined }
 import eu.timepit.refined.auto._
+import eu.timepit.refined.boolean.Not
 import eu.timepit.refined.numeric._
 import eu.timepit.refined.scalacheck.numeric._
 import eu.timepit.refined.util.time.Minute
 import org.scalacheck.Prop._
 import org.scalacheck.Properties
+
+import shapeless.Nat
 
 class NumericArbitrarySpec extends Properties("NumericArbitrary") {
 
@@ -17,11 +20,17 @@ class NumericArbitrarySpec extends Properties("NumericArbitrary") {
   property("Less.negative") =
     checkArbitraryRefType[Refined, Int, Less[W.`-100`.T]]
 
+  property("Less.natural") =
+    checkArbitraryRefType[Refined, Long, Less[Nat._10]]
+
   property("LessEqual.positive") =
     checkArbitraryRefType[Refined, Int, LessEqual[W.`42`.T]]
 
   property("LessEqual.negative") =
     checkArbitraryRefType[Refined, Int, LessEqual[W.`-42`.T]]
+
+  property("LessEqual.natural") =
+    checkArbitraryRefType[Refined, Long, LessEqual[Nat._10]]
 
   property("Greater.positive") =
     checkArbitraryRefType[Refined, Int, Greater[W.`100`.T]]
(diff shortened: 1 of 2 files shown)


jedesah:
There is still a few things to add. In particular an extra property test, as it stands this will probably decrease coverage, but wanted to open the PR sooner rather than later in order to get feedback and avoid extra work.

fthomas:
Thanks for the PR, @jedesah. I'll take a deeper look at it later today.

contrib/scalacheck/shared/src/test/scala/eu/timepit/refined/scalacheck/NumericArbitrarySpec.scala
```diff
@@ -17,24 +20,45 @@ class NumericArbitrarySpec extends Properties("NumericArbitrary") {
...
     checkArbitraryRefType[Refined, Int, Greater[W.`100`.T]]
 
   property("Greater.negative") =
     checkArbitraryRefType[Refined, Int, Greater[W.`-100`.T]]
 
+  property("Greater.natural") =
+    checkArbitraryRefType[Refined, Long, Greater[Nat._10]]
+
+  property("Greater range") =
+    forAll { (min: Refined[Long, Greater[Nat._10]], max: Refined[Long, Greater[Nat._10]]) =>
```
fthomas: If `min > max` `greaterArbitraryRange[Refined, Long, Nat._10](min, max)` probably doesn't produce any values which seems to be the cause of the failed Travis build.

fthomas:
The `Bounded` type class is appealing but it currently doesn't give us an `Arbitrary[BigInt Refined Positive]`, right? One way to "fix" this would be to add a low priority implicit `Bound[T]` from any `nt: Numeric[T]` with `def maxValue = nt.fromInt(Int.MaxValue)` and `def minValue = nt.fromInt(Int.MinValue)`. What do you think?

jedesah:
I think it might be better to provide a (low priority?) instance for `BigInt` which is `Int.MaxValue` or `Long.MaxValue`. I think it's dangerous to assume that an unknown numeric type will support `Int.MaxValue`. I would rather the user code get a compile error asking for an instance of the `Bounded` typeclass than a runtime error because `nt.fromInt(Int.MaxValue)` throws an exception.

Btw, what are the different types for which I need to create instances? Probably all the types in the standard library that have instances for `Numeric`?

contrib/scalacheck/shared/src/test/scala/eu/timepit/refined/scalacheck/NumericArbitrarySpec.scala
jedesah: Yup, that seems like a reasonable hypothesis. Can probably add a precondition on the outer property.

jedesah: Tried that. The filter is still triggering scalacheck to abandon. I think there is a way to increase the ratio of discarded examples without triggering the property to abandon the "proof". Or is there a better way to encode information in the types?

fthomas: You could do a runtime check which of `min` and `max` is the smaller resp. greater value and then pass the smaller as first arg and the greater as second arg to `greaterArbitraryRange`.

fthomas:
If I'm handed a `nt: Numeric[T]` I would never expect that `nt.fromInt(x)` throws an exception for any `x`. You would also not expect that `Numeric[T].compare` throws exceptions, right? But I'm also fine with just an instance for `BigInt`.

Yes, we should have instances for all types that have `Numeric` instances in the std lib. A list of these is http://www.scala-lang.org/api/current/index.html#scala.math.Numeric under "Known Subclasses".

jedesah:
Ah, I assumed it would throw an exception, but I just tested it out with `Short` and it overflows, so `Numeric[Short].fromInt(Int.MaxValue) == -1`. But that does not seem a whole lot better. Arguably it's even worse than throwing an exception as it could fail in silent or hard to diagnose ways.

fthomas:
I also just learned that `fromInt` is a cast in disguise. So adding a `Bounded[BigInt]` instance seems to be the better alternative.

PR #108, opened 2015-12-15
Add RefType.refineMF

### LINKED ISSUE #107: Support aliasing RefineM
The `RefineM` macro can currently be invoked via the `refineM` functions (e.g. `refineMV[Positive](1)`) or via an implicit conversion which needs to be imported with `import auto._` (it then can be invoked with `1: (Int Refined Positive)`). It would be nice if users of refined could write their own aliases for `RefineM` with fixed types and without importing an implicit conversion. See https://github.com/quasar-analytics/quasar/pull/1025 for more context.

We can support this with a variation of the `RefineMPartiallyApplied` class:
```scala
final class RefineMFullyApplied[F[_, _], T, P] {
  def apply(t: T)(implicit v: Validate[T, P], rt: RefType[F]): F[T, P] = macro RefineM.macroImpl[F, T, P]
}

val Natural = new RefineMFullyApplied[Refined, Long, NonNegative]
```

```scala
scala> Natural(1L)
res3: eu.timepit.refined.api.Refined[Long,eu.timepit.refined.numeric.NonNegative] = Refined(1)

scala> Natural(-1)
<console>:47: error: Predicate (-1 < 0) did not fail.
       Natural(-1)
              ^

scala> Natural(1.3)
<console>:47: error: type mismatch;
 found   : Double(1.3)
 required: Long
       Natural(1.3)
               ^
```

/cc @jedesah

fthomas:
`refineMF` is a variation of `refineM` but which requires that the base type is specified before `apply` can be called. This allows to create aliases like `val Natural = RefType[Refined].refineMF[Long, NonNegative]` which only can be called with `Long` arguments.

closes #107

/cc @jedesah and @wemrysi

### DIFF
--- a/core/shared/src/main/scala/eu/timepit/refined/api/RefType.scala
+++ b/core/shared/src/main/scala/eu/timepit/refined/api/RefType.scala
@@ -1,7 +1,7 @@
 package eu.timepit.refined
 package api
 
-import eu.timepit.refined.internal.{ ApplyRefPartiallyApplied, RefineMPartiallyApplied, RefinePartiallyApplied }
+import eu.timepit.refined.internal._
 import shapeless.tag.@@
 
 import scala.reflect.macros.blackbox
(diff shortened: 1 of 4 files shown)


jedesah:
:+1:

fthomas:
I'm not particular fond of the name `refineMF`, but other than that I like this change. It complements `refineM` nicely. If someone has a better suggestion for the name, I'm all ears.

PR #109, opened 2015-12-16
Arbitrary instances for numeric predicates indexed by Nats

fthomas:
This expands on #106 by using four helper functions to implement all numeric `Arbitrary` instances independent of the predicates being indexed by literal singleton types or `Nat`s. I think that is the best we can do to avoid duplication in those definitions.

Thanks to @jedesah for the initial implementation!

@jedesah I hope you don't mind moving this forward. I think it is in your interest to get this and #108 released ASAP.

### DIFF
--- a/contrib/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/numeric.scala
+++ b/contrib/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/numeric.scala
@@ -1,7 +1,7 @@
 package eu.timepit.refined
 package scalacheck
 
-import eu.timepit.refined.api.RefType
+import eu.timepit.refined.api.{ RefType, Validate }
 import eu.timepit.refined.numeric._
 import org.scalacheck.Gen.Choose
 import org.scalacheck.{ Arbitrary, Gen }
@@ -43,6 +43,18 @@ object numeric {
     implicit val short: Bounded[Short] =
       Bounded(Short.MinValue, Short.MaxValue)
   }
+
+  /**
+   * A generator that generates a random value in the given (inclusive)
+   * range that satisfies the predicate `P`. If the range is invalid,
+   * the generator will not generate any value.
+   */
+  def chooseRefinedNum[F[_, _], T: Numeric: Choose, P](min: F[T, P], max: F[T, P])(
+    implicit
+    rt: RefType[F],
+    v: Validate[T, P]
+  ): Gen[F[T, P]] =
+    Gen.chooseNum(rt.unwrap(min), rt.unwrap(max)).filter(v.isValid).map(rt.unsafeWrap)
 
   ///
 
--- a/contrib/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/numeric.scala
+++ b/contrib/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/numeric.scala
@@ -3,7 +3,6 @@ package scalacheck
 
 import eu.timepit.refined.api.RefType
 import eu.timepit.refined.numeric._
-import eu.timepit.refined.boolean.Not
 import org.scalacheck.Gen.Choose
 import org.scalacheck.{ Arbitrary, Gen }
 import shapeless.ops.nat.ToInt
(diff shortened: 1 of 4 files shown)


jedesah:
@fthomas Of course not! :smile: Thx for picking up where I left off!

contrib/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/numeric.scala
```diff
@@ -3,92 +3,165 @@ package scalacheck
...
-    nt: Numeric[T],
-    c: Gen.Choose[T]
+    wh: Witness.Aux[H]
   ): Arbitrary[F[T, Interval.Closed[L, H]]] =
-    arbitraryRefType(Gen.chooseNum(wl.value, wh.value))
+    rangeClosedArbitrary(wl.value, wh.value)
+
+  ///
+
+  private def rangeOpenArbitrary[F[_, _]: RefType, T: Choose, P](min: T, max: T)(
```
jedesah: Do you think we could put all the instances of `Arbitrary` in a trait and then we could make these functions protected? I am asking, because then I could re-use them in order to implement `greaterArbitraryRange(min,max)`.

Btw, if you are wondering why I bother with `greaterArbitraryRange(min,max)` instead of simply making the types of input to properties more precise, it's because I like having "pure" properties and then supplying a generator at the bottom that does not actually generate huge numbers when dealing with properties for which that would make the property execute in an unreasonable amount of time. This way there is a separation between the "pure" expression of logical property and the practical considerations of "verifying" it using sample value generation.

fthomas: We can also make them public, I've no problem with that.

I think I now better understand the purpose of `greaterArbitraryRange`. That is something you would use explicitly when writing a property, right? So you would write `property("xyz") = forAll(greaterArbitraryRange(Natural(0), Natural(10)).arbitrary) { n => ??? }`. I think it should be then a `Gen` instead of an `Arbitrary`.

Maybe we can also have a variant of `greaterArbitraryRange` that supports all numeric predicates:
```scala
def closedRangeWithBounds[F[_, _], T: Numeric: Choose, P](min: F[T, P], max: F[T, P])(
  implicit rt: RefType[F]
): Gen[F[T, P]] =
  Gen.choose(rt.unwrap(min), rt.unwrap(max)).map(rt.unsafeWrap)
```
This _should_ work with predicates that are indexed by literal singleton types and `Nat`s. Usage would then look like this
```scala
forAll(closedRangeWithBounds(Natural(0), Natural(10))) { n =>
  ???
}
```

fthomas: The implementation of `closedRangeWithBounds` is too simple. `P` could exclude values in the range `[min, max]`, for example if `P = IsEven`. So we also need to filter with a `Validate[T, P]` there.

jedesah: [Here](https://github.com/quasar-analytics/quasar/pull/1025/files#diff-f9a99a7d09f0195334ae06d9b3e366a2R125) is an example usage in quasar.

Specs2 needs it to be in the form of an `Arbitrary` unless I am missing some alternate syntax. But you may be right that possibly semantically it makes more sense for it to be a Generator. I am not 100% clear on the semantic differences between `Arbitrary` and `Gen`.

So I think we should keep the existing methods private, because they allow you to "cheat" and I think that's fine because it keeps the verbosity level minimal internally and expose the new public function that you proposed which actually validates that the `min` and `max` are in the domain of the refinement. Too bad we can't verify that `max` > `min` which was the cause of the issue you pointed out earlier.

fthomas: One difference between `Arbitrary` and `Gen` is that the former is mostly used as context bound and the latter is asked for explicitly. That matches your usage in quasar/specs2 where e.g. `greaterArbitraryMax` (which is an `Arbitrary`) is used in an `implicit` parameter list. So I think it makes sense to make it a `Gen`.

Ok, so I've added the function I proposed as `chooseRefinedNum`. I thought about adding `greaterArbitraryMax` but those cannot be defined as generically as `chooseRefinedNum`. I could imagine that it is cleaner to use `chooseRefinedNum` instead, although you'd have to pass the other bound explicitly.

jedesah: :+1:

PR #124, opened 2016-01-12
Remove workaround that was introduced for #3

fthomas:


fthomas:
The workaround is still needed for 2.11. Interestingly enough it is not needed for 2.10.

PR #135, opened 2016-03-10
Add Adjacent type class

fthomas:
This adds an `Adjacent` type class to `refined-scalacheck` to make the numeric `Arbitrary` instances more robust. The type class allows to get rid of `filter` calls in the numeric generators.


fthomas:
Btw, this PR was inspired by https://github.com/quasar-analytics/quasar/pull/1102

PR #140, opened 2016-03-21
Add new boolean predicate Nand

sh0hei:


fthomas:
This looks good to me, thanks @sh0hei! Have you considered defining `Nand` as a combination of `And` and `Not`, i.e.
```scala
type Nand[A, B] = Not[A And B]
```
?
This definition wouldn't require an associated `Validate` instance. The only downside is that `showExpr` is less clear:
```scala
scala> type Nand[A, B] = Not[A And B]
defined type alias Nand

scala> 5: Int @@ Nand[Positive, Greater[_1]]
<console>:49: error: Predicate ((5 > 0) && (5 > 1)) did not fail.
       5: Int @@ Nand[Positive, Greater[_1]]
       ^
```

sh0hei:
Hi, @fthomas !

thank you for your comment.

I'm understood to be able to express `Nand` by `And` and `Not`.

But `Xor` which can be expressed by `And` and `Not` equally existed, so `Nand` was added.

But I think simplicity is fine for an expression of `type Nand[A, B] = Not[A And B]` as you say.

I'll Close this PR

Thanks :)

fthomas:
@sh0hei , qou're of course right, that we could express `Xor` in terms of other predicates. That would probably look like this:
```scala
type Xor[A, B] = Not[Not[A And Not[A And B]] And Not[B And Not[A And B]]]
```
:smile:

I think the clarity of the current `showExpr` for `Xor` would beat the `showExpr` of this definition of `Xor`. So it is a trade-off between readable messages but more code and less code but less readable messages.

I'm fine with merging this PR if you'd change `Nand` to be an alias as outlined above.

sh0hei:
Hi, @fthomas :)


> I'm fine with merging this PR if you'd change Nand to be an alias as outlined above.

Thank you very much :+1: 


I'll update this PR, Thanks in advance for your help :)

fthomas:
That looks great, thanks @sh0hei!

sh0hei:
Thank you, too :)

PR #146, opened 2016-03-31
Specialize RefType.unsafeWrap

fthomas:
This is an experiment to test how `@specialized` affects `RefType.unsafeWrap`.


fthomas:
This is not needed anymore due to #149.

PR #156, opened 2016-05-02
Use scalafmt

fthomas:


core/shared/src/test/scala/eu/timepit/refined/BooleanValidateSpec.scala
```diff
@@ -4,18 +4,18 @@ import eu.timepit.refined.TestUtils._
...
 class BooleanValidateSpec extends Properties("BooleanValidate") {
 
-  type FF[Op[_, _]] = False Op False
-  type FT[Op[_, _]] = False Op True
-  type TF[Op[_, _]] = True Op False
-  type TT[Op[_, _]] = True Op True
+  type FF [Op[_, _]] = False Op False
+  type FT [Op[_, _]] = False Op True
+  type TF [Op[_, _]] = True Op False
+  type TT [Op[_, _]] = True Op True
```
fthomas: @olafurpg This looks also strange. There shouldn't be a space between the name of a type and its parameter list.

olafurpg: There should be a space, this is fixed in the already released 0.2.2 but because that release introduced https://github.com/olafurpg/scalafmt/issues/192 I haven't advertised it.

olafurpg: btw, the type signature here doesn't fit on 80 characters

```diff
+  implicit def lowerCaseArbitrary[F[_, _]: RefType]: Arbitrary[F[
+          Char, LowerCase]] = arbitraryRefType(Gen.alphaLowerChar)
```

I'm sure `--maxColumn 100` will yield better results.

fthomas:
Closing in favor of another PR that uses a newer scalafmt version.

PR #157, opened 2016-05-06
Add support for play-json

lunaryorn:
@fthomas There you go :)

Couldn't get SBT to exclude the play code from 2.10 builds, though… any idea?  Maybe it's better to turn this into a completely separate project?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,6 +29,7 @@ val shapelessVersion = "2.3.0"
 val scalaCheckVersion = "1.12.5"
 val scalazVersion = "7.2.2"
 val scodecVersion = "1.9.0"
+val playVersion = "2.5.0"
 
 val allSubprojects = List("core", "scalacheck", "scalaz", "scodec")
 val allSubprojectsJVM = allSubprojects.map(_ + "JVM")
@@ -46,7 +47,8 @@ lazy val root = project.in(file("."))
     scalazJVM,
     scalazJS,
     scodecJVM,
-    scodecJS)
+    scodecJS,
+    play)
   .settings(commonSettings)
   .settings(noPublishSettings)
   .settings(releaseSettings)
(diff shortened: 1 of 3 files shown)


fthomas:
Thanks @lunaryorn!

I pushed your changes to https://github.com/fthomas/refined/tree/play-support and will try to exclude 2.10 there.

fthomas:
I still need to figure out how I can integrate [sbt-doge](https://github.com/sbt/sbt-doge) into the build and if it plays well with the release plugin. Sorry for the delay!

lunaryorn:
@fthomas Never mind. I fought sbt-doge for a while, but eventually gave up.  My SBT foo is too bad.

If it's too much hassle I can just as well make this a separate project and publish independently 😊

fthomas:
Play itself uses sbt-doge und sbt-release and requires [this kind of customisation](https://github.com/playframework/playframework/blob/58a150d712bdf3bf4d14709569afcfd9ea97fb7f/framework/project/Release.scala#L34-L54) so that both plugins work together. That scares me and I'd like to avoid making refined's build even more complicated than it is right now. I also haven't found any other project that combines both plugins in a simple way.

So at this point, I think it would be easier for both of us to put this into a separate project.

lunaryorn:
@fthomas Fine with me.  I'll see if I get a repo up tonight.  Maybe I'll need your help for the release, I've never published any Scala artefacts to a public repository all by myself, no clue how to do that :blush:

fthomas:
Great! Ping me on Gitter if you need any help. Maybe I remember what needs to be done to publish to Sonatype / Maven Central. :smile:

fthomas:
Let's close this since https://github.com/lunaryorn/play-json-refined is now published. Thanks again!

lunaryorn:
@fthomas You're welcome 😊 was a lot of fun, and I learned a bit along the way 😎

PR #158, opened 2016-05-06
Document why Refined is not a value class for 2.10

fthomas:
/cc @dwijnand


dwijnand:
I see.. kinda. That message is confusing

Issue #57, opened 2015-08-26
Overwrite Refined's copy


fthomas:
https://twitter.com/FouriersTrick/status/636539026208608257 and https://issues.scala-lang.org/browse/SI-7884

TL;DR case class copy method subverts the private constructor

REPL session:
```scala
scala> val r = refineMV[Positive](5)
r: eu.timepit.refined.Refined[Int,eu.timepit.refined.numeric.Positive] = Refined(5)

scala> r.copy[Int, Positive](-5)
res5: eu.timepit.refined.Refined[Int,eu.timepit.refined.numeric.Positive] = Refined(-5)
```
:-(

Thanks @ceedubs for the PSA.

fthomas:
After pondering about this more, I came to the conclusion that I don't wont to fix this. SI-7884 is a scalac bug that needs to be addressed but I don't want a workaround for it in refined. It is just one of the escape hatches in Scala and if people want to workaround the guarantees refined provides they can always use `Refined.unsafeApply` or `asInstanceOf`. Making `copy` private pollutes refined more that it provides any more safety.

PR #160, opened 2016-05-06
Drop case in Refined

dwijnand:
.. with before-and-after (newly created) spec changes

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -162,7 +162,15 @@ lazy val submoduleJvmSettings = Seq(
     Seq(
       exclude[DirectMissingMethodProblem  ]("eu.timepit.refined.generic.evalValidate"),
       exclude[DirectMissingMethodProblem  ]("eu.timepit.refined.GenericValidate.evalValidate"),
-      exclude[ReversedMissingMethodProblem]("eu.timepit.refined.GenericValidate.evalValidate")
+      exclude[ReversedMissingMethodProblem]("eu.timepit.refined.GenericValidate.evalValidate"),
+      exclude[MissingTypesProblem]("eu.timepit.refined.api.Refined"),
+      exclude[DirectMissingMethodProblem]("eu.timepit.refined.api.Refined.productElement"),
+      exclude[DirectMissingMethodProblem]("eu.timepit.refined.api.Refined.productArity"),
+      exclude[DirectMissingMethodProblem]("eu.timepit.refined.api.Refined.canEqual"),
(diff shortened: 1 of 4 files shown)


dwijnand:
Btw, happy to iterate and tweak any minor detail, the intent of the change is my only goal.

dwijnand:
Let me know what you want to do in terms of bincompat..

fthomas:
MiMa complains about a lot of stuff that `case` brought in. Of those things, I think I only care about `scala.Serializable`. What do you think of the others?

dwijnand:
I'm happy to recover `Serializable`.

The other one that's a shot-call on your part - do you want to provide `Refined.unapply`?

fthomas:
Yes, `Refined.unapply` would be useful too.

dwijnand:
Thanks to @ceedubs it seems this was already decided against in #57. Perhaps this alternative fix is more acceptable.

dwijnand:
@fthomas Added `Refined.unapply`, squashed and pushed.

fthomas:
Thanks @dwijnand! LGTM. So the only unsafe operations are now `Refined.unsafeApply` and `asInstanceOf`, right?

dwijnand:
`Refined.unsafeApply` - yes. What do you mean by `asInstanceOf`? Can you show a code/repl snippet?

fthomas:
```scala
scala> val x: Int Refined Positive = 5
x: eu.timepit.refined.api.Refined[Int,eu.timepit.refined.numeric.Positive] = 5

scala> x.asInstanceOf[Int Refined Negative]
res0: eu.timepit.refined.api.Refined[Int,eu.timepit.refined.numeric.Negative] = 5
```

dwijnand:
Right.. yeah, that one too :)

dwijnand:
Custom value classes don't have that problem.

fthomas:
I'm wondering if we should make `Refined` a case class again by making it a [`sealed abstract case class`](https://gist.github.com/tpolecat/a5cb0dc9adeacc93f846835ed21c92d2). With this combination of keywords we don't get unsafe `apply` and `copy` methods.

dwijnand:
Can't have sealed value classes, as value classes are final.

fthomas:
Good catch, @dwijnand. I've heard that the value class aspect of `Refined` was a reason for not using refined in the first place since value classes can't be nested. Maybe `extends AnyVal` should be dropped from `Refined`.

dwijnand:
Maybe value classes should be allowed to nest... :D

fthomas:
I heard that is an easy change in scalac ;-)

PR #179, opened 2016-06-15
Add scalaz.{Equal, Show} instances

fthomas:


fthomas:
BTW, since `Equal` and `Show` are both contravariant functors, the two instances could be implemented in terms of a general function that abstracts over type classes that are contravariant functors:
```scala
  implicit def refTypeEqual[F[_, _]: RefType, T: Equal, P]: Equal[F[T, P]] =
    xxx[F, T, P, Equal]

  implicit def refTypeShow[F[_, _]: RefType, T: Show, P]: Show[F[T, P]] =
    xxx[F, T, P, Show]

  def xxx[F[_, _], T, P, TC[_]: Contravariant](implicit rt: RefType[F], tc: TC[T]): TC[F[T, P]] =
    tc.contramap(rt.unwrap)
```

PR #193, opened 2016-07-26
WIP: Type class that combines RefType and Validate  

fthomas:
This adds the `RefinedType` type class that combines `RefType` and `Validate` instances for a refined type `F[T, P]`. The motivating example for this type class was given by @benhutchison on [Gitter](https://gitter.im/fthomas/refined?at=5781c8edc9b49c1d6ff03072). Abstracting over `RefType.applyRef` currently requires a lot of type ceremony:

``` scala
def refineEff[S] = new ApplyRefPartiallyAppliedExt[S]

final class ApplyRefPartiallyAppliedExt[S] {
  def apply[E: Err, F[_, _], T, P](t: T)(
    implicit ev: F[T, P] =:= S, rt: RefType[F], v: Validate[T, P]
  ): Eff[E, S] =
    fromXor(RefType.applyRef[S].apply(t).toXor)
}
```

The goal is to allow calls like `refineEff[PosInt](5)` where `PosInt` is an alias for `Int Refined Positive` and the base type `Int` can be inferred from the parameter of `apply`. With `RefinedType`, this can be simplified to:

``` scala
def refineEff[S] = new ApplyRefPartiallyAppliedExt[S]

final class ApplyRefPartiallyAppliedExt[S] {
  def apply[E: Err, T](t: T)(
    implicit rt: RefinedType.AuxT[S, T]
  ): Eff[E, S] =
    fromXor(rt.refine(t).toXor)
}
```

What has improved with `RefinedType`:
- `apply` has two fewer type parameters
- `apply` has two fewer implicit parameters
- `RefType.applyRef[S].apply(t)` became `rt.refine(t)`

In general I think this is a win. The only downside is that we still require the `ApplyRefPartiallyAppliedExt` class to partially apply the refined type `S` and to let the base type `T` be inferred.

Also the combination of `RefType` and `Validate` is very common (e.g., type class instances of 3rd-party libraries for refined types almost always require the combination of these two), so it is natural to combine them. It also makes the implementation of refined simpler in some cases.

It is unfortunate that the names `RefType` and `RefinedType` are very similar. This type class makes me regret the name `RefType`. Maybe something like `CarrierType` or `WrapperType` would have been a better choice.


benhutchison:
+1. Much simpler. Im keen to try it out in a project!

fthomas:
Superseded by #369 which also contains the `RefinedType` type class.

PR #209, opened 2016-10-19
Build with 2.12.0-RC2

fthomas:
This currently fails because shapeless 2.3.2 is not yet available for 2.12.0-RC2.


fthomas:
This now fails because scodec-core is not yet available for RC2.

PR #251, opened 2017-01-21
Move compiler-dependent code into a module

### LINKED ISSUE #207: Core depends on scala-compiler
The stuff in `generic` is cool. Like super-cool. However, a great deal of the time, you don't need the tremendous power that it provides for defining arbitrary tests. When this is the case, it's really a drag that pulling in refined pulls in `scala-compiler.jar`, as `scala-compiler.jar` is _huge_ (like 15mb). I know it's really a hassle to split projects up into modules, but it would really be great if `generic` was in its own module in order to shrink the disk-wise footprint of refined's dependency set.

clhodapp:
Fixes #207
I rejiggered the parts of the build related to binary dependency checking a bit and bumped the version to 0.7.0 (-SNAPSHOT), since the change is breaking.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,7 +22,8 @@ val pureconfigVersion = "0.5.0"
 val macroParadise = compilerPlugin(
   "org.scalamacros" % "paradise" % macroParadiseVersion % Test cross CrossVersion.full)
 
-val allSubprojects = Seq("core", "scalacheck", "scalaz", "scodec", "pureconfig")
+val allSubprojects =
+  Seq("core", "eval", "scalacheck", "scalaz", "scodec", "pureconfig")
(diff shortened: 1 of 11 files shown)


fthomas:
This looks great, @clhodapp. Thanks for taking care of this issue!

clhodapp:
Thanks for merging! A release including this would be highly appreciated sooner rather than later, as I'm finding that I get classloading failures under Scala 2.12 with Typelevel Scala (that would be fixed by a release of this change).

clhodapp:
Also :crossed_fingers: that I didn't break the release process. I _think_ it should work.

fthomas:
I'll try to release 0.7.0 today or tomorrow. One thing I'd like to do before is to mark the dependency on `scala-reflect` also as "provided".

PR #253, opened 2017-01-24
Mark the dependency on scala-reflect as "provided"

fthomas:
see also https://github.com/fthomas/refined/issues/207


fthomas:
Hmm, this failed. Not sure why this works for shapeless or circe but not for refined. Anyhow, closing.

PR #254, opened 2017-01-25
Fix build

fthomas:


clhodapp:
D'oh! I guess I changed it from `Seq` to `Set` and forgot to update the template... Sorry about that! Perhaps this will work now, as the artifacts are now in maven central.

fthomas:
No worries! The build still fails because refined-eval is not yet on Maven Central. Since refined goes to Maven Central via Bintray, new modules need to be added to JCenter before they can be synced to Maven Central. I'm still waiting for Bintray to add refined-eval to JCenter.

Issue #260, opened 2017-02-28
nondeterministic compiler crash ¯\_(ツ)_/¯


tpolecat:
I saw this on Travis, compiling doobie at commit b2f02078d0162827b93a1d3408c2a9950ce477d4 under ++2.11.8. I restarted the build and it worked fine, so it's nondeterministic. I'm offering this on the chance that you will find it useful, but I don't expect a fix so feel free to close.

```
[error] /home/travis/build/tpolecat/doobie/modules-cats/refined/target/scala-2.11/src_managed/test/scala/doobie/refined/refinedtypes.scala:82: exception during macro expansion: 
[error] java.lang.AssertionError: assertion failed: import failure: cannot determine unique overloaded method alternative from
[error]  final package eu
[error] final package eu
[error]  that matches package eu:eu.type
[error] 	at scala.Predef$.assert(Predef.scala:186)
[error] 	at scala.reflect.internal.Importers$StandardImporter.disambiguate$1(Importers.scala:181)
[error] 	at scala.reflect.internal.Importers$StandardImporter.recreateOrRelink$1(Importers.scala:188)
[error] 	at scala.reflect.internal.Importers$StandardImporter.importSymbol(Importers.scala:210)
[error] 	at scala.reflect.internal.Importers$StandardImporter.recreateOrRelink$1(Importers.scala:169)
[error] 	at scala.reflect.internal.Importers$StandardImporter.importSymbol(Importers.scala:210)
[error] 	at scala.reflect.internal.Importers$StandardImporter.recreateOrRelink$1(Importers.scala:166)
[error] 	at scala.reflect.internal.Importers$StandardImporter.importSymbol(Importers.scala:210)
[error] 	at scala.reflect.internal.Importers$StandardImporter.recreateOrRelink$1(Importers.scala:166)
[error] 	at scala.reflect.internal.Importers$StandardImporter.importSymbol(Importers.scala:210)
[error] 	at scala.reflect.internal.Importers$StandardImporter.recreatedTreeCompleter(Importers.scala:298)
[error] 	at scala.reflect.internal.Importers$StandardImporter$$anonfun$importTree$1.apply$mcV$sp(Importers.scala:417)
[error] 	at scala.reflect.internal.Importers$StandardImporter$$anonfun$tryFixup$1.apply(Importers.scala:49)
[error] 	at scala.reflect.internal.Importers$StandardImporter$$anonfun$tryFixup$1.apply(Importers.scala:49)
[...]

PR #262, opened 2017-03-06
Improve MacroUtils.eval

fthomas:
by calling `duplicate` before untypechecking the tree.

This is an attempt to fix https://github.com/fthomas/refined/issues/260


fthomas:
BTW, thanks to @btre for making me aware of `duplicate`.

PR #264, opened 2017-03-07
Change refined-pureconfig to use pureconfig 0.6.0

vlovgr:
Outstanding questions:
- ~~We need to populate the `toTyp` parameter of `CannotConvert` with a `String` representation of `F[T, P]`. I don't see a straightforward way to do this without using type tags, and we would then lose any type aliases. Adding the type tag breaks binary compatibility (according to MiMa).~~
- ~~`PredicateFailedException` is now unused but has not been removed as that would break binary compatibility (according to MiMa).~~

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -221,9 +221,7 @@ lazy val moduleJvmSettings = Def.settings(
   mimaBinaryIssueFilters ++= {
     import com.typesafe.tools.mima.core._
     import com.typesafe.tools.mima.core.ProblemFilters._
-    Seq(
-      exclude[DirectMissingMethodProblem](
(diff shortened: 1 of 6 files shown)


vlovgr:
@fthomas: I think this is ready for you to have a look at. :)

fthomas:
I'm okay with breaking bincompat and bumping the version to 0.8.0 for removing `PredicateFailedException` and adding a `TypeTag` to `refTypeConfigConvert`. I guess pureconfig also uses `TypeTag` for populating the `toTyp` parameter?

vlovgr:
@fthomas I had a look at PureConfig and both it and some of it's modules rely on class tags, so we should be fine using type tags here. I've done both changes, so this should be ready from my side!

PR #279, opened 2017-04-29
Update to Scala.js 0.6.16

fthomas:


fthomas:
Thanks to https://github.com/scala-js/scala-js/pull/2908 this allows to delete our own `def nextAfter(start: Float, direction: Double): Float`.

PR #284, opened 2017-05-30
Show and Eq instances for cats

derekmorr:
This adds cats `Show` and `Eq` instances for refined types, similar to the scalaz module.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,7 +65,6 @@ lazy val cats = crossProject
     libraryDependencies += "org.typelevel" %%% "cats-core" % catsVersion,
     initialCommands += s"""
       import $rootPkg.cats._
-      import $rootPkg.cats.auto._
     """
   )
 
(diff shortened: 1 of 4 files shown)


derekmorr:
Thanks for the mima fix. Let me know if you want me to squash the commit so there isn't a broken build in the history.

build.sbt
```diff
@@ -55,6 +58,20 @@ lazy val root = project
...
   )
 
+lazy val cats = crossProject
+  .configureCross(moduleCrossConfig("cats"))
+  .dependsOn(core % "compile->compile;test->test")
+  .settings(
+    libraryDependencies += "org.typelevel" %%% "cats-core" % catsVersion,
+    initialCommands += s"""
+      import $rootPkg.cats._
+      import $rootPkg.cats.auto._
```
fthomas: There is no `cats.auto` object in this PR. Running `cats/console` will fail with an error, right?

derekmorr: Ah, yes, my mistake. I'll update the PR.

PR #291, opened 2017-06-04
Example usge of "fully refined type" (Type + Predicate)

ronanM:
```scala
// Fully refined type: type (Int) with predicate (here: 7 <= value <= 77).
type Age = Int Refined Interval.ClosedOpen[W.`7`.T, W.`77`.T]

val userInput = 55

// With typed value declaration.
val ageEither1 : Either[String, Age] = refineV(userInput)
println(ageEither1)

// With inferred type value.
val ageEither2 = RefType.applyRef[Age](userInput)
println(ageEither2)
```


fthomas:
LGTM, thanks!

PR #293, opened 2017-06-06
Update PureConfig URL

leifwickland:
PureConfig managed to snag the `pureconfig` org on github, so its official location shifted.


fthomas:
Thanks!

Issue #294, opened 2017-06-13
Add more PortNumber types


fthomas:
#249 added the `PortNumber` type for `Int`s in the range from 0 to 65535. Recently several talks / blog posts demoed a type for non-system ports (1024 to 65535):

* https://underscore.io/blog/posts/2017/03/07/refined-data-config-database.html
* https://vlovgr.github.io/refined-types/#16
* http://leifwickland.github.io/presentations/configBomb/#66

So I think it makes sense to add at least a `NonSystemPortNumber` type. https://en.wikipedia.org/wiki/Registered_port suggests we could also have `SystemPortNumber`, `UserPortNumber`, and `PrivatePortNumber`. But I  guess these would be used less frequently.

fthomas:
Fixed by #304

PR #304, opened 2017-07-04
Add more port-number types

sh0hei:
Implementation for Issue https://github.com/fthomas/refined/issues/294


fthomas:
This is great, many thanks. I've only one remark, can we remove the `WellKnownPortNumber`, `RegisteredPortNumber`, and `PrivatePortNumber` aliases? Having one name for each range should be enough.

PR #314, opened 2017-08-14
Add support for decline

alonsodomin:
Adds derivation of [decline](http://ben.kirw.in/decline/)'s `Argument` classes for defining refined types for command line arguments.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,6 +19,7 @@ val scalazVersion = "7.2.15"
 val scodecVersion = "1.10.3"
 val pureconfigVersion = "0.7.2"
 val jsonpathVersion = "2.3.0"
+val declineVersion = "0.3.0"
 
 // needed for tests with Scala 2.10
 val macroParadise = compilerPlugin(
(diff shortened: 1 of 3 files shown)


alonsodomin:
got this to a point in which it does build for the _correct_ Scala versions but fails for the Scala 2.10 build because decline has not been published for that Scala version.

This makes me wonder whether it would be better to PR this in the decline side instead, thoughts?

alonsodomin:
This has been PR'ed as bkirwi/decline#12 and build passes there. If it gets accepted there, we can close this one.

fthomas:
This looks great @alonsodomin! Technically it should be possible to build subprojects with different `crossScalaVersions` using [sbt-doge](https://github.com/sbt/sbt-doge) but I've struggled with it before (see https://github.com/fthomas/refined/pull/157). So I think it would be easier for everyone if this is added to decline directly. :-)

alonsodomin:
@fthomas SBT 1.0 fixes the problems with `crossScalaVersions` but still... I agree, it even _looks_ to me like the _natural_ direction of the dependency should be from decline into refined...

alonsodomin:
PR against decline has been merged so this can be closed. happy days.

Issue #328, opened 2017-09-22
Reduce compilation times


fthomas:
This tweet https://twitter.com/missingfaktor/status/909064863917510656 shows that increased compilation time with refined is a major issue. We should investigate ways to get better in this regard.

fthomas:
The first step towards this goal has been merged: ~26% reduced compilation times with #332.

fthomas:
Another step in this direction: #333

fthomas:
General performance improvements: #335, #336, #337, and #338

fthomas:
Reduced compilation time when refining with full refined types: #388

PR #330, opened 2017-09-25
Specialized macros for autoRefine{V,T}

fthomas:
This specializes the macro implementations for `autoRefineV` and
`autoRefineT`. Both methods were defined in terms of the generic
`RefineMacro.impl` which works for any `RefType[F]`. This generic
implementation has the drawback that it needs to `eval` the implicit
`RefType` parameter at compile-time. Since `autoRefine{V,T}` already
know which `RefType` instance they will use, it is unnecessary to `eval`
them in the macro.

Here are the benchmark results for master and this change:
```
master:

[info] Result "eu.timepit.refined.benchmark.RefineMacroBenchmark.autoRefineV_PosInt":
[info]   0.549 ±(99.9%) 0.004 s/op [Average]
[info]   (min, avg, max) = (0.520, 0.549, 0.593), stdev = 0.015
[info]   CI (99.9%): [0.546, 0.553] (assumes normal distribution)
[info] # Run complete. Total time: 00:17:45
[info] Benchmark                                Mode  Cnt  Score   Error Units
[info] RefineMacroBenchmark.autoRefineV_PosInt  avgt  200  0.549 ± 0.004 s/op

impl_Refined:

[info] Result "eu.timepit.refined.benchmark.RefineMacroBenchmark.autoRefineV_PosInt":
[info]   0.403 ±(99.9%) 0.004 s/op [Average]
[info]   (min, avg, max) = (0.377, 0.403, 0.471), stdev = 0.017
[info]   CI (99.9%): [0.399, 0.407] (assumes normal distribution)
[info] # Run complete. Total time: 00:09:07
[info] Benchmark                                Mode  Cnt  Score   Error Units
[info] RefineMacroBenchmark.autoRefineV_PosInt  avgt  200  0.403 ± 0.004 s/op
```

That is a decrease in compilation time of ~26%. This is first PR towards
reduced compilation times: https://github.com/fthomas/refined/issues/328

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/auto.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/auto.scala
@@ -64,9 +64,8 @@ object auto {
    * This is an implicit version of `[[refineMV]]`.
    */
   implicit def autoRefineV[T, P](t: T)(
-      implicit rt: RefType[Refined],
-      v: Validate[T, P]
-  ): Refined[T, P] = macro RefineMacro.impl[Refined, T, P]
+      implicit v: Validate[T, P]
+  ): Refined[T, P] = macro RefineMacro.impl_Refined[T, P]
 
   /**
    * Implicitly tags (at compile-time) a value of type `T` with `P` if `t`
@@ -76,7 +75,6 @@ object auto {
    * This is an implicit version of `[[refineMT]]`.
    */
   implicit def autoRefineT[T, P](t: T)(
-      implicit rt: RefType[@@],
-      v: Validate[T, P]
-  ): T @@ P = macro RefineMacro.impl[@@, T, P]
+      implicit v: Validate[T, P]
+  ): T @@ P = macro RefineMacro.impl_@@[T, P]
 }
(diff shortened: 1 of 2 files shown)


soronpo:
AFAIK, `eval` won't be part of the future macro system, so this is a good step towards forward compatibility

fthomas:
@soronpo I agree. `eval` is slow and error-prone so I will not complain if it is going away. But there is another use of `eval` in the refine macro that will be much harder to replace.

soronpo:
> But there is another use of eval in the refine macro that will be much harder to replace.

Yes. That is why singleton-ops has a large case statement and a recursive calculation of the type tree..

fthomas:
@soronpo The "disadvantage" of that approach is that such macro cannot use third-party code. Everything singleton-ops can do at compile-time must be defined in singleton-ops.

fthomas:
Closing, since this PR has been superseded by #332.

soronpo:
>  The "disadvantage" of that approach is that such macro cannot use third-party code. Everything singleton-ops can do at compile-time must be defined in singleton-ops.

Yes, indeed. But composition of basic operation can take you pretty far.

PR #332, opened 2017-09-27
Do not use eval for known RefType instances

fthomas:
This avoids to call `eval` for known `RefType` instances in
`RefineMacro.impl`. This is similar to #330 but more general
since it is an optimization for all macros in the library and
not only for `autoRefine{V,T}`.

The performance improvements are similar to that of #330:
```
[info] Result "eu.timepit.refined.benchmark.RefineMacroBenchmark.autoRefineV_PosInt":
[info]   0.408 ±(99.9%) 0.003 s/op [Average]
[info]   (min, avg, max) = (0.387, 0.408, 0.442), stdev = 0.012
[info]   CI (99.9%): [0.405, 0.411] (assumes normal distribution)
[info] # Run complete. Total time: 00:09:11
[info] Benchmark                                Mode  Cnt  Score   Error Units
[info] RefineMacroBenchmark.autoRefineV_PosInt  avgt  200  0.408 ± 0.003 s/op

```

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/macros/RefineMacro.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/macros/RefineMacro.scala
@@ -1,10 +1,11 @@
 package eu.timepit.refined
 package macros
 
-import eu.timepit.refined.api.{RefType, Validate}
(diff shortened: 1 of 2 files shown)


fthomas:
Another advantage over #330 is that this does not change any public API.

PR #343, opened 2017-10-18
Split objects in the scalacheck module into traits for easier reuse.

nrinaudo:
This is also more coherent with the style used [here](https://github.com/fthomas/refined/tree/master/modules/core/shared/src/main/scala/eu/timepit/refined/types), for example.


modules/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/all.scala
```diff
@@ -0,0 +1,10 @@
+package eu.timepit.refined.scalacheck
+
+object all
+    extends AnyInstances
```
fthomas: I'm wondering if we should mix `AnyInstances` into `all`. The instance there is able to provide an `Arbitrary` for any refined type. I could imagine that this could cause ambiguous implicit errors when importing `all._`.

nrinaudo: I thought that this wouldn't be an issue since implicit resolution rules would always give priority to more specific types, but maybe I'm mistaken.

Another reason not to include would be the high rate of of discarded instances.

Fixed in 89f45cad98f0ca2c42cbc977fbfb7aaba1d671ab.

fthomas: You're probably right there about instances with more specific types.

The high rate of discarded values makes it useless in many cases which is the better reason for not mixing it into `all`.

Issue #36, opened 2015-07-20
Predicates for BigInt and BigDecimal


fthomas:

fthomas:
Closing. These could still be useful but let's add them when they are requested.

shokohara:
I want this function.

fthomas:
@shokohara Unfortunately I hadn't described in details what I meant with this issue, but I think the general idea was to have predicates for `String` which are valid arguments for the `scala.math.BigInt(s: String)` and `scala.math.BigDecimal(s: String)` constructors, for example something like this:
```scala
String Refined ThisIsAValidBigInt
```
Are we taking about the same thing here?

kusamakura:
@fthomas That looks useful, but I'm also running into cases where it would be nice to have BigInts/BigDecimals constructed from literals not result in the `compile-time refinement only works with literals` compile-time error.

fthomas:
@kusamakura This is possible and would just need a small change in `RefineMacro`. [This pattern match](https://github.com/fthomas/refined/blob/b83acba91563ab0a29a2d78f4bebfac6ab10d647/modules/core/shared/src/main/scala/eu/timepit/refined/macros/RefineMacro.scala#L21-L24) would need additional cases for trees that look like `BigInt(<literal>)` and `BigDeciaml(<literal>)`. A PR for that would be very much welcomed. :-)

kusamakura:
@fthomas Thanks for pointing me in the right direction. I think I kind of got it: it works for "normal" cases. I'm quite new to macros, so please excuse the bad code.

```scala
val tValue: T = t.tree match {
  case Literal(Constant(value)) =>
    value.asInstanceOf[T]
  case q"scala.`package`.BigInt.apply(${lit: Literal})" =>
    BigInt(lit.value.value.asInstanceOf[Int]).asInstanceOf[T]
  case q"scala.`package`.BigDecimal.apply(${lit: Literal})" =>
    BigDecimal(lit.value.value.asInstanceOf[Double]).asInstanceOf[T]
  case _ =>
    abort(Resources.refineNonCompileTimeConstant)
}
```

Needless to say, for expressions like `val m: BigInt Refined Positive = BigInt(0x7fffffffffffffffL)`, I get a `ClassCastException`. I'm not sure how I can match on possible types of the literal so it can be passed as an argument to BigInt/BigDecimal constructors. Additionally, that line of thought kind of intersects with your idea for `String Refined ThisIsAValidBigInt` as well.

[...]

PR #345, opened 2017-10-22
Refined literals support for BigInt and BigDecimal

kusamakura:
From #36.

It should work fine for numeric predicates except `Equals`.
There seems to be some difficulty with obtaining a Validate instance for BigInt and BigDecimal.


fthomas:
Looks great @kusamakura! Could you run `sbt scalafmt` and commit to fix the build failure?

kusamakura:
@fthomas Sorry for the delay. I'm thinking of trying to make `Equals` work with BigInt and BigDecimal first, but just in case that that would be for another issue, I'll push fixes to the build failure.

fthomas:
It seems that `Unliftable` is not available on 2.10, so you can't use that. And the "pattern var ... in method unapply is never used" warning in the 2.12 build seems to be bogus. I'm okay with commenting out the `"-Ywarn-unused:patvars"` option in `build.sbt` which should get rid of this warning.

kusamakura:
@fthomas Sorry, I wasn't able to work on this last week. The previous commit builds fine now, but I'm not sure how to go about the code coverage error. (I don't know how it works, sorry.)

kusamakura:
@fthomas I've read up on code coverage, but I'm not sure if scoverage can detect matchers in macros. It seems to do so since `RefineMacro.scala` is covered, but I can't seem to get it to see the matchers `LiteralMatchers.scala` being used in the tests I provided.

fthomas:
@kusamakura I think the problem with the code coverage is caused by using `scala-2.11+` as directory component for `BigLiteralsSpec.scala`. Looking at the build log for 2.12 I can't find the string `BigLiterals` indicating that this test is not executed in the 2.12 build. AFAIK sbt does not support directories like `scala-2.11+` out of the box so you need to duplicate this test file for 2.11 and 2.12.

kusamakura:
@fthomas That makes sense. Thank you so much for your patience! I'll push the changes in a jiffy.

kusamakura:
Thanks again @fthomas. I learned a lot. :)

Issue #342, opened 2017-10-18
Provide companion objects with helper methods for value creation


nrinaudo:
One barrier to the adoption of refined is that it's not trivial to create values at runtime - one needs to know that the type is refined, and how to work with these.

Ideally, types would be self contained, and value creation methods would be provided in a companion object.

For example:

```scala
type NonEmptyString = String Refined NonEmpty

// Some macro magic would generate:
object NonEmptryString {
  def fromString(str: String): Either[String, NonEmptyString] =
    RefType.applyRef[NonEmptyString](str)
}
```

Ideally, there's also be an `apply` method for literal values.

This would let library authors, for example, provide powerful types using refined without exposing users to the underlying implementation details.

nrinaudo:
Here's an improvement on that idea. You could also add extractors. In the previous example:

```scala
object NonEmptyString {

  def fromString(str: String): Either[String, NonEmptyString] =
    RefType.applyRef[NonEmptyString](str)

  def unapply(str: String): Option[String] = fromString(str).toOption

}
```

Some use cases:
```scala
// Filtering
List("", "1", "abc").collect {
  case NonEmptyString(s) => s
}

// Guards + extractors
val s1: Option[String] = ...
val s2: Option[String] = ...

for {
  NonEmptyString(str1) <- s1
  NonEmptyString(str2) <- s2
} yield str1 + str2

// ...
```

fthomas:
I really like this idea and started working on a macro for this in #346. It works with Scala 2.12 but unfortunately not with 2.10 and with Scala.js on 2.11. :-(

fthomas:
Since my effort to do this with a macro failed horribly, I started working on a non-macro solution in #369 which I'm quite happy with and which will probably be in the next release.

nrinaudo:
honestly, I provably like this better than a macro solution.

nrinaudo:
For the record, it's a pattern I use quite a bit (see [this](https://github.com/nrinaudo/kantan.codecs/blob/develop/core/src/main/scala/kantan/codecs/Decoder.scala#L128), for example).

PR #346, opened 2017-10-25
WIP: Macro that generates companion for refined types

fthomas:
This is the second draft of a macro for #342 and still WIP.


fthomas:
The current implementation works with 2.12 but not with 2.10 and 2.11. And unfortunately both fail in different ways and I've no idea how to fix them.

fthomas:
Closing in favor of #369.

PR #349, opened 2017-10-28
Move list of projects using refined back to README

fthomas:
This list currently lives in the wiki where (I think) it is mostly ignored. Moving it back to the README gives projects that provide integrations with refined much more attention.


NeQuissimus:
Big 👍 from me for this :)

PR #353, opened 2017-11-13
Add string checks for IPv4, IPv6

NeQuissimus:
Uses `java.net.Inet{,4,6}Address` to determine whether a String is a valid IPv4/IPv6.

### DIFF
--- a/modules/core/jvm/src/test/scala/eu/timepit/refined/StringValidateSpecJvm.scala
+++ b/modules/core/jvm/src/test/scala/eu/timepit/refined/StringValidateSpecJvm.scala
@@ -6,6 +6,21 @@ import org.scalacheck.Prop._
 import org.scalacheck.Properties
 
 class StringValidateSpecJvm extends Properties("StringValidate") {
+  property("IPv4.isValid") = secure {
+    isValid[IPv4]("10.0.0.1")
+  }
+
+  property("IPv4.showResult.Failed") = secure {
+    showResult[IPv4]("::1") ?= "IPv4 predicate failed: requirement failed"
+  }
+
+  property("IPv6.isValid") = secure {
+    isValid[IPv6]("::1")
+  }
+
+  property("IPv6.showResult.Failed") = secure {
+    showResult[IPv6]("10.0.0.1") ?= "IPv6 predicate failed: requirement failed"
+  }
 
   property("Regex.showResult") = secure {
     showResult[Regex]("(a|b") ?=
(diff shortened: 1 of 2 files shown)


NeQuissimus:
What should I do about the MiMa errors about 2.11 and 2.10?
I was not expecting additions to cause binary compatibility issues...

fthomas:
@NeQuissimus In principle these predicates would be a welcome addition but I'm afraid we can't use [`java.net.InetAddress.getByName`](https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#getByName-java.lang.String-) in the `Validate` instances because of their side-effecting nature. It seems that this method performs a DNS lookup if the given string is a hostname and only checks the validity of the address format if a literal IP address is provided. This means that code like
```scala
val x: String Refined IPv4 = "github.com"
```
would perform a DNS lookup at compile-time and therefore compilation would be dependent on the network. This should be avoided.

As far as I can see `java.net.InetAddress` has no function that just checks the format of literal IP addresses. Fortunately there is a [`refined-guava`](https://github.com/derekmorr/refined-guava/blob/master/src/main/scala/refined/guava/net/net.scala) module by @derekmorr that provides `IPv4` and `IPv6` predicates (besides others) which uses Guava's `com.google.common.net.InetAddresses.forString` function which deliberately avoids all DNS lookups.

NeQuissimus:
@fthomas Oooh, I did not know about `refined-guava`! It does exactly what I need then :)
I realized just today that it would allow DNS lookups, so I was already planning to update this PR... I'll just close it now.

fthomas:
Out of curiosity, is there an easy way to implement these predicates without adding additional dependencies?

NeQuissimus:
Well, you could create Regex' to match them but then you'll end up with giant regular expressions :)
I mean, look at this: https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses/17871737

All the JDK stuff makes DNS calls as far as I know.

NeQuissimus:
Actually, we could boolean a very simply regex on top of this PR to avoid DNS lookups...

derekmorr:
I can look into published refined-guava. That was based on some work I did that also needed IP address parsing, but I didn't want to side-effect. Unfortunately, guava is a big dependency for that.

@fthomas Would you want this as a module for refined, or would you prefer it as a separate project?

NeQuissimus:
@derekmorr Or, we could use apache/commons-validator, which is much smaller, and would allow us to add a bunch of stuff. (think credit card numbers, email address...)
I'd be more than willing to help out with this (since I will be using it :D)

fthomas:
I don't have any experience with commons-validator but if it is more suitable than Guava, go for it. However, one downside of Java dependencies is that we won't be able to cross-compile it for Scala.js and Scala Native.
A new module in refined for these predicates is fine with me.

PR #356, opened 2017-11-17
Add IPv4 and private networks

NeQuissimus:
Basically a follow-up to #353

This is what I am using at work, I figured I would clean it up and contribute it back.

No external Java dependencies, so this also works with Scala.js without any issues.

I will have to add IPv6 next week, so I will send that in as well once I have it.

/cc @derekmorr @fthomas

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/string.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/string.scala
@@ -15,6 +15,9 @@ object string extends StringValidate with StringInference {
   /** Predicate that checks if a `String` ends with the suffix `S`. */
   final case class EndsWith[S](s: S)
 
+  /** Predicate that checks if a `String` is a valid IPv4 */
+  final case class IPv4()
+
   /** Predicate that checks if a `String` matches the regular expression `S`. */
   final case class MatchesRegex[S](s: S)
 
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/main/scala/eu/timepit/refined/string.scala
```diff
@@ -40,14 +43,34 @@ object string extends StringValidate with StringInference {
   final case class XPath()
 }
 
-private[refined] trait StringValidate {
+object Predicates {
+  object IPv4 {
```
fthomas: Could you move the `IPv4` object into the `string` object right below the `IPv4` case class and remove the `Predicates` object? It makes sense to put auxiliary code of predicates into their proper companion.

modules/core/shared/src/main/scala/eu/timepit/refined/types/net.scala
```diff
@@ -23,4 +25,61 @@ trait NetTypes {
...
+
+  /** A `String` representing a valid IPv4 in the private network 172.15.0.0/12 (RFC1918) */
+  type Rfc1918ClassBPrivate = String Refined Rfc1918ClassBPrivateSpec
+
+  /** A `String` representing a valid IPv4 in the private network 192.168.0.0/16 (RFC1918) */
+  type Rfc1918ClassCPrivate = String Refined Rfc1918ClassCPrivateSpec
+
+  /** A `String` representing a valid IPv4 in a private network according to RFC1918 */
+  type Rfc1918Private =
+    String Refined Rfc1918ClassAPrivateSpec Or Rfc1918ClassBPrivateSpec Or Rfc1918ClassCPrivateSpec
```
fthomas: Would it make sense to add an alias `type Rfc1918PrivateSpec = Rfc1918ClassAPrivateSpec Or Rfc1918ClassBPrivateSpec Or Rfc1918ClassCPrivateSpec`? That could be used here and in the definition of `PrivateNetwork`. A `type Rfc5737TestnetSpec` could be also useful to avoid repetition.

fthomas:
Looks great, many thanks @NeQuissimus!

PR #357, opened 2017-11-22
Add IPv6 string validation

NeQuissimus:
Looks like IPv6 validation using a regex is what makes the most sense.

- I looked at `sun.net.util.IPAddressUtil` but that's an internal JDK API, not supposed to be used, not available for Scala.JS (it is the underlying implementation for `java.net.InetAddress`)
- `java.net.InetAddress` itself always tries to make DNS calls unless you convert the String into the address bytes. That would, however, require a lot of logic to undo compacting etc.
- Guava and Apache Commons Validator are JVM-only and fairly large

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/string.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/string.scala
@@ -33,6 +33,44 @@ object string extends StringValidate with StringInference {
     }
   }
 
+  /** Predicate that checks if a `String` is a valid IPv6 */
+  final case class IPv6()
+  object IPv6 {
+    val ipv4Chars = "(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])"
+    val ipv4 = s"(${ipv4Chars}\\.){3,3}${ipv4Chars}"
+    val ipv6Chars = "[0-9a-fA-F]{1,4}"
+
+    val ipv6Full = s"(${ipv6Chars}:){7,7}${ipv6Chars}" // 1:2:3:4:5:6:7:8
+
+    val ipv6Compact = List(
+      s"(${ipv6Chars}:){1,7}:", // 1:: .. 1:2:3:4:5:6:7::
+      s"(${ipv6Chars}:){1,6}:${ipv6Chars}", // 1::8 .. 1:2:3:4:5:6::8
+      s"(${ipv6Chars}:){1,5}(:${ipv6Chars}){1,2}", // 1::7:8 .. 1:2:3:4:5::8
+      s"(${ipv6Chars}:){1,4}(:${ipv6Chars}){1,3}", // 1::6:7:8 .. 1:2:3:4::8
+      s"(${ipv6Chars}:){1,3}(:${ipv6Chars}){1,4}", // 1::5:6:7:8 .. 1:2:3::8
+      s"(${ipv6Chars}:){1,2}(:${ipv6Chars}){1,5}", // 1::4:5:6:7:8 .. 1:2::8
+      s"(${ipv6Chars}:)(:${ipv6Chars}){1,6}", // 1::3:4:5:6:7:8 .. 1::8
+      s":((:${ipv6Chars}){1,7}|:)" // ::2:3:4:5:6:7:8 .. ::
+    )
+
+    val interface = "[0-9a-zA-Z]{1,}"
+    val ipv6LinkLocal = s"fe80:(:${ipv6Chars}){0,4}%${interface}"
+
+    val mappedIpv6 = s"::(ffff(:0{1,4}){0,1}:){0,1}${ipv4}"
+
+    val embeddedIpv4 = s"(${ipv6Chars}:){1,4}:${ipv4}"
+
+    val formats = List(
+      ipv6Full,
+      ipv6LinkLocal,
+      mappedIpv6,
+      embeddedIpv4
+    ) ++ ipv6Compact
+
+    val regex = formats.map(regex => s"(^${regex}$$)").mkString("|").r.pattern
+    val predicate: String => Boolean = s => regex.matcher(s).matches
+  }
+
   /** Predicate that checks if a `String` matches the regular expression `S`. */
   final case class MatchesRegex[S](s: S)
 
@@ -82,6 +120,9 @@ private[refined] trait StringValidate {
 
   implicit def ipv4Validate[S <: String]: Validate.Plain[String, IPv4] =
     Validate.fromPredicate(IPv4.predicate, t => s"${t} is a valid IPv4", IPv4())
+
+  implicit def ipv6Validate[S <: String]: Validate.Plain[String, IPv6] =
+    Validate.fromPredicate(IPv6.predicate, t => s"${t} is a valid IPv6", IPv6())
 
   implicit def matchesRegexValidate[S <: String](
       implicit ws: Witness.Aux[S]): Validate.Plain[String, MatchesRegex[S]] =
--- a/build.sbt
+++ b/build.sbt
@@ -272,7 +272,9 @@ lazy val moduleJvmSettings = Def.settings(
       ProblemFilters.exclude[MissingClassProblem]("eu.timepit.refined.scalacheck.util.OurMath$"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("eu.timepit.refined.StringValidate.*"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("eu.timepit.refined.types.*"),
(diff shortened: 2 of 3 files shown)


modules/core/shared/src/main/scala/eu/timepit/refined/string.scala
```diff
@@ -15,6 +15,14 @@ object string extends StringValidate with StringInference {
   /** Predicate that checks if a `String` ends with the suffix `S`. */
   final case class EndsWith[S](s: S)
 
+  /** Predicate that checks if a `String` is a valid IPv6 */
+  final case class IPv6()
+  object IPv6 {
+    val regex =
+      "^(^(([0-9A-Fa-f]{1,4}(((:[0-9A-Fa-f]{1,4}){5}::[0-9A-Fa-f]{1,4})|((:[0-9A-Fa-f]{1,4}){4}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,1})|((:[0-9A-Fa-f]{1,4}){3}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,2})|((:[0-9A-Fa-f]{1,4}){2}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,3})|(:[0-9A-Fa-f]{1,4}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,4})|(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,5})|(:[0-9A-Fa-f]{1,4}){7}))$|^(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,6})$)|^::$)|^((([0-9A-Fa-f]{1,4}(((:[0-9A-Fa-f]{1,4}){3}::([0-9A-Fa-f]{1,4}){1})|((:[0-9A-Fa-f]{1,4}){2}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,1})|((:[0-9A-Fa-f]{1,4}){1}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,2})|(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,3})|((:[0-9A-Fa-f]{1,4}){0,5})))|([:]{2}[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,4})):|::)((25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{0,2})\\.){3}(25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{0,2})$$".r.pattern
```
fthomas: Can we factor out common subexpressions to reduce the length of that regex? `[0-9A-Fa-f]{1,4}` seems to be a good candidate on a first glance. I'd imagine doing it like this:
```scala
val regex = {
  val group = "[0-9A-Fa-f]{1,4}"
  s"^(^(($group(((:$group){5}::$group) ..."
}
```

derekmorr:
java.net.InetAddress only makes DNS calls if the input isn't an IP address literal. IPv6 address parsing is pretty complicated. It doesn't look like this parser handles embedded IPv4-literals:

```scala

scala>     val regex =
     |       "^(^(([0-9A-Fa-f]{1,4}(((:[0-9A-Fa-f]{1,4}){5}::[0-9A-Fa-f]{1,4})|((:[0-9A-Fa-f]{1,4}){4}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,1})|((:[0-9A-Fa-f]{1,4}){3}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,2})|((:[0-9A-Fa-f]{1,4}){2}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,3})|(:[0-9A-Fa-f]{1,4}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,4})|(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,5})|(:[0-9A-Fa-f]{1,4}){7}))$|^(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,6})$)|^::$)|^((([0-9A-Fa-f]{1,4}(((:[0-9A-Fa-f]{1,4}){3}::([0-9A-Fa-f]{1,4}){1})|((:[0-9A-Fa-f]{1,4}){2}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,1})|((:[0-9A-Fa-f]{1,4}){1}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,2})|(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,3})|((:[0-9A-Fa-f]{1,4}){0,5})))|([:]{2}[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,4})):|::)((25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{0,2})\\.){3}(25[0-5]|2[0-4][0-9]|[0-1]?[0-9]{0,2})$$".r.pattern
regex: java.util.regex.Pattern = ^(^(([0-9A-Fa-f]{1,4}(((:[0-9A-Fa-f]{1,4}){5}::[0-9A-Fa-f]{1,4})|((:[0-9A-Fa-f]{1,4}){4}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,1})|((:[0-9A-Fa-f]{1,4}){3}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,2})|((:[0-9A-Fa-f]{1,4}){2}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,3})|(:[0-9A-Fa-f]{1,4}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,4})|(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,5})|(:[0-9A-Fa-f]{1,4}){7}))$|^(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,6})$)|^::$)|^((([0-9A-Fa-f]{1,4}(((:[0-9A-Fa-f]{1,4}){3}::([0-9A-Fa-f]{1,4}){1})|((:[0-9A-Fa-f]{1,4}){2}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,1})|((:[0-9A-Fa-f]{1,4}){1}::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,2})|(::[0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,3})|((:[0-9A-Fa-f]{1,4}){0,5})))|([:]{2}[0-9A-Fa-f]{1,4}(:[0-9A-Fa-...

scala> regex.matcher("2001:db8:122:344::192.0.2.33").matches // example from https://tools.ietf.org/html/rfc6052#section-2.4
res1: Boolean = false

scala> import java.net.InetAddress // But the JDK's parser works fine
import java.net.InetAddress

scala> InetAddress.getByName("2001:db8:122:344::192.0.2.33")
res2: java.net.InetAddress = /2001:db8:122:344:0:0:c000:221
```

fthomas:
Thanks @derekmorr for this test case. https://stackoverflow.com/a/17871737/460387 shows that a regex can handle embedded IPv4 literals. @NeQuissimus what do you think about updating the regex to also support embedded IPv4 literals?

NeQuissimus:
Sorry, I completely missed the feedback here. I will take a look.
I was aware of this, my regex only validates against RFC-8200.
I'll see if I can't make some improvements or maybe find a better regex online.

kusamakura:
It might also be a nice touch to have an `Inference` for `IPv4 ==> IPv6`, if that's possible. (Or also the other way around for valid values.)

NeQuissimus:
I fixed up the regex to account for embedded IPv4 addresses and I also extracted common elements.
I added tests for the new address formats.

@fthomas Can you kick off the Travis build again? No matter how often I run the tests locally, they pass...

fthomas:
The Scala.js tests failed with
```
[info] ! StringValidate.IPv6.isValid.compact: Falsified after 0 passed tests.
```
This should be reproducible with `coreJS/test`.

NeQuissimus:
wow, this took me a while! JavaScript's Regex engine has different eagerness rules than the one in the JVM...
I fixed it now and hopefully Travis will agree...

Issue #363, opened 2017-12-08
Validate instances for ByteVector with collection predicates


fthomas:
The `refined-scodec` module could provide `Validate` instances for `ByteVector` with collection predicates. @coltfred suggested on Gitter that `Size[P]` would be useful predicate to use with `ByteVector`. I'd like to explore if it is possible to write such instances.

fthomas:
This has been done in #365.

PR #365, opened 2017-12-09
Collection validate instances for ByteVector

fthomas:
This is still WIP and for now only contains an instance for `Size`.

Example usage:
```scala
scala> refineV[Size[Equal[W.`2L`.T]]](ByteVector.fromValidHex("0xabcd"))
res0: Either[String, Refined[ByteVector, Size[Equal[Long(2L)]]]] = Right(ByteVector(2 bytes, 0xabcd))

scala> refineV[Size[Equal[W.`2L`.T]]](ByteVector.fromValidHex("0xabcdef"))
res1: Either[String, Refined[ByteVector, Size[Equal[Long(2L)]]]] = Left(Predicate taking size(ByteVector(3 bytes, 0xabcdef)) = 3 failed: Predicate failed: (3 == 2).)
```

This implements #363. /cc @coltfred

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -196,7 +196,11 @@ lazy val scodec = crossProject(JSPlatform, JVMPlatform)
     libraryDependencies ++= Seq(
       "org.scodec" %%% "scodec-core" % scodecVersion,
       macroParadise
-    )
+    ),
+    initialCommands += s"""
+      import $rootPkg.scodec.predicates.byteVector._
+      import _root_.scodec.bits.ByteVector
+    """
   )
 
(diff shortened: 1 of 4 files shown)


coltfred:
This is awesome! Thanks! I was going to offer to write tests, but by the time I got the chance you'd already written them. I'd happily start using this tomorrow if I could, even with just Sized.

fthomas:
Ok, great. I'm going to merge this PR as is. If there is a need for supporting more predicates or adding support for `BitVector`, we can always add that later.

This will be in version 0.8.5 which will be released shortly after Cats 1.0.0 is released.

PR #369, opened 2017-12-15
Add RefinedTypeOps for companion objects of refined types

fthomas:
This is a solution for #342 but unlike #346 it does not add a new macro. Compared to a macro solution, companion objects need be written manually and require a little bit boilerplate. E.g. for a refined type `type PosInt = Int Refined Positive` one would need to write a companion that extends from `RefinedTypeOps` like this:
```scala
object PosInt extends RefinedTypeOps[PosInt, Int]
```
Usage then looks like:
```scala
scala> PosInt(5)
res4: PosInt = 5

scala> PosInt(-5)
<console>:51: error: Predicate failed: (-5 > 0).
       PosInt(-5)
             ^

scala> PosInt.from(5)
res6: Either[String,PosInt] = Right(5)

scala> val PosInt(x) = 5
x: PosInt = 5
```

So this PR adds the `RefinedTypeOps` class to simplify the definition of companion objects of refined types, the `RefinedType` type class that combines `RefType` and `Validate` instances of a refined type (thus making #193 obsolete), and companion objects for refined types in the `eu.timepit.refined.types` package.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -261,14 +261,14 @@ lazy val moduleJvmSettings = Def.settings(
   mimaBinaryIssueFilters ++= {
     import com.typesafe.tools.mima.core._
     Seq(
-      ProblemFilters.exclude[MissingClassProblem]("eu.timepit.refined.scalacheck.util.OurMath"),
-      ProblemFilters.exclude[MissingClassProblem]("eu.timepit.refined.scalacheck.util.OurMath$"),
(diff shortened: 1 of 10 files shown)


fthomas:
Another advantage of this approach over a macro that just adds some functions to an object is that we can easily add extension methods to any refined type companion. See #382 for details.

PR #376, opened 2017-12-19
Add `Order` instance to cats module

howyp:
Delegates to the `Order` instance for the wrapped type, in the same way as the `Eq`/`Show` instances


howyp:
I have no idea why https://github.com/fthomas/refined/pull/376/commits/0a08bb6e70f293ea013ed6014aea712f063622c1 didn't compile. Anyway, should be good to go now

PR #379, opened 2017-12-20
Valid number string predicates

kusamakura:
Adds the following string predicates: `ValidInt`, `ValidLong`, `ValidDouble`, `ValidBigInt`, and `ValidBigDecimal`.

I initially implemented this with a typeclass-based approach, e.g. `ParsableString[N]`, which the `Validate` instance would use, but it seemed to be double the work than just creating new predicates and the corresponding `Validate` instances. Both seem extensible enough, though I surmise that a typeclass-based approach would be a little less verbose, since supplying the typeclass instances would be enough to get started. @fthomas, if you think that that's a more sensible approach, then I can always update this PR.

I'm also thinking of adding inference instances for these, but I'll hold off for now until I can find a good way of doing it without copious amounts of copy and paste.


fthomas:
Thanks @kusamakura, this looks great!

The `ParseableString[N]` approach is an interesting idea. Instead of defining a predicate and the `Validate` instance you just need to define the `ParseableString` instance. But as you already noted, the latter is just a little bit less work and probable a little bit harder to understand.

Issue #75, opened 2015-10-17
Add error accumulating function to RefType


fthomas:
Add a function to `RefType` that is similar to `refine` but instead of returning an `Either` it should return a type that can accumulate errors (something like [`cats.Validated`](http://non.github.io/cats//tut/validated.html)). bond has this as can be seen here: https://github.com/fthomas/refined/blob/v0.8.2/modules/docs/feature_comparison_bond.md#what-bond-has-but-refined-not

fthomas:
https://meta.plasm.us/posts/2015/12/17/error-accumulating-decoders-in-circe/ might be interesting here.

ysusuk:
hi @fthomas, i'll be glad to work on this issue) can you give me some hints on where to start?

ysusuk:
hi @fthomas 
did you mean as well to create another `RefinePartiallyApplie` in cats package, that will return ValidatedNel instead of Either?

fthomas:
@ysusuk Yes. :-)

ysusuk:
@fthomas can you pls assign this issue on me)

fthomas:
It seems that I only can assign it to myself.

fthomas:
I'm closing this in favor of #382.

PR #382, opened 2017-12-20
Add .validate method for refined companions

howyp:
Now we have 'companion' objects for refined types, I thought it would be nice to have a neat way to refine to cat's `Validated` type.

It always uses `NonEmptyList` on the error side at the moment. I guessed that this is what people will use 90% time but if we think there will also be need for using `List` or even `List Refined Not[Empty]` then I we could support that too?


fthomas:
Looks great, thanks @howyp!

This would also resolve #75 although that one is about adding a method to `RefType`. My guess is that companions of refined types will replace the `RefType.applyRef`, `refineV` and similar methods in the long run so it is okay to have this functionality only on the companion objects.

modules/cats/shared/src/test/scala/eu/timepit/refined/cats/CatsSpec.scala
```diff
@@ -11,16 +12,26 @@ class CatsSpec extends Properties("cats") {
     val refTypeOrder: Unit = () // shadow the `Order` instance so the `Eq` instance is tested
     locally(refTypeOrder) // prevent a "unused" warning
 
-    PosInt.unsafeFrom(5) === PosInt.unsafeFrom(5)
```
fthomas: I'm using `unsafeFrom` in the tests to avoid the overhead of the `RefineMacro` so that test code compiles faster.

modules/cats/shared/src/main/scala/eu/timepit/refined/cats/CatsValidated.scala
```diff
@@ -0,0 +1,11 @@
+package eu.timepit.refined.cats
+
+import cats.data.{NonEmptyList, Validated, ValidatedNel}
+import eu.timepit.refined.api.RefinedTypeOps
+
+object validation {
+  implicit class CatsValidateOps[FTP, T](companion: RefinedTypeOps[FTP, T]) {
+    def validate(value: T): ValidatedNel[String, FTP] =
+      Validated.fromEither(companion.from(value)).leftMap(NonEmptyList.one)
```
fthomas: There is a `toValidatedNel` on Cats' `EitherSyntax` which would allow us to just write `companion.from(value).toValidatedNel` here.

fthomas:
I'm totally fine with using `ValidatedNel` here. It is probably the predominant data type for error accumulation in the Cats ecosystem.

modules/cats/shared/src/test/scala/eu/timepit/refined/cats/CatsSpec.scala
howyp: Ah, ok, I'll revert

howyp:
Not sure about packaging here - do we need a `.syntax` as well as a =`.validation` one? Might be overkill

fthomas:
I'm not sure if I like d59f2d8ab235179e99e36ce2ffc8fc25d97ec347. It adds a `validate` extension to any type and I find such extensions always strange. I would be okay if we only supply the `validate` method for companions and that is the only way to refine values with `Validated`.
I know that the library currently has more than one way to do things and `RefType` is one of the worst contender in this regard. I think in the future I'd like to reduce the surface area of refined's API to make the library simpler.

Regarding `syntax` vs. `validation`. I think I like `syntax` more than `validation`. It would allow us to maybe add more extension methods that are not related to `Validated` to the companions.

kusamakura:
How about some scalaz love? ;) I can take it up if you want, @howyp.

fthomas:
Adding this functionality to the Scalaz module would be nice too. I would prefer a separate PR for that.

kusamakura:
@fthomas Alright, I'll graft the code from this one onto the scalaz module.

howyp:
I can understand why you might feel hesitant about having an operator on anything (though it would only work on types which have a matching refined predicate). It seemed to me to be quite a 'cats' way of doing things - in the style of say [`.valid[B]`](https://typelevel.org/cats/api/cats/syntax/ValidatedIdSyntax.html). The alternative is to support something like:

```scala
import validation._
val myInt = 5
validate[Positive](myInt)
```
Would you prefer that for the moment?

My motivation was to provide symetrical support for cases where you haven't defined a full refined type/don't want to create companion objects, but just want to be able to apply a predicate to a raw type. I agree it would be good to distill the API surface of Refined, and that standardising on companion objects for refined types is a great way to go. But only providing second-class support for ad-hoc refinments via a predicate seems a shame to me.

Perhaps how to reduce the surface of the API would be best discussed in a new issue?

kusamakura:
@howyp I think `validate[Positive](myInt)` would just be `refineV[Positive](myInt).toValidatedNel`. On the other hand, the companion method is nice.

howyp:
Ok, I've dropped the `10.validate[Positive]` syntax

fthomas:
Ok, merging now. Thanks again @howyp!

Issue #184, opened 2016-07-06
Some form of über-import to access standard functions in single line


benhutchison:
Some people (myself among them) prefer the simplicity of a blanket import to enable standard functionality, along the lines of `import scalaz._; import Scalaz._`.

 As well as convenience and accessibility for newcomers ("just make it work"!), the other reason Ive come to appreciate them is I like to detect name clashes between libs I use early.

fthomas:
I think it would make sense if the über-import would include everything in the `refined` package object, all predicates (that is `boolean._` , `char._`, `collection._`, `generic._`, `numeric._`, and `string._`), `auto._`, `types.all._`, and `api.Refined`. This should include everything that is needed for "normal" usage of the library. Everything else is only needed if one wants to extend the library or define their own predicates.

soronpo:
Maybe it can be divided into severl über-imports, classified according to use.

- Normal usage (`import refined.normal`) - What is stated above.
- Advanced usage (`import refined. advanced`) - Normal + predicate extenders 
- Others... I don't know all the use cases, but I believe you can narrow them down.

fthomas:
One step in this direction has been made in #433 which is now available in 0.9.0. So to import all predicates we can now `import eu.timepit.refined.predicates.all._`.

To import all of the standard functionality of the library we are now down to five imports:
```scala
import eu.timepit.refined._
import eu.timepit.refined.api.Refined
import eu.timepit.refined.auto._
import eu.timepit.refined.predicates.all._
import eu.timepit.refined.types.all._
```
My long-term goal is to shrink that further to maybe two or three imports.

PR #383, opened 2017-12-21
Add Min/Max operations for numeric types

howyp:
I'd quite like for there to be a way to get and use the minimum/maximum values of a refined type.

This is a *very* early WIP for adding a `Min` typeclass and implementation, inspired by @fthomas's `InhabitantsOf`. I think it's viable but wanted to make sure everyone was happy with the concept/implementation approach before I continue?

Operations such as `.all`/`InhabitantsOf` could be derived from the min/max for numerics. Also, I'd guess we could re-work the scalacheck arbitrary instances to derive from the `Min` and `Max` typeclasses.


howyp:
This now has support for both `.min` and `.max`, for Integer types with most numeric predicates.

Not yet supported:
- `Float`/`Double` base types
- `Modulo` predicate

Both may be tricky.

howyp:
I've just seen `eu.timepit.refined.scalacheck.util.Bounded` - there's probably some commonality to be removed here

fthomas:
@howyp I like the idea and approach of this PR. The symmetry between base types and refined types is nice (e.g.: `Int.MaxValue <-> PosInt.max`).

howyp:
Hmm... I'm starting to think that `PosInt.MaxValue` would be better than `PosInt.max`, to have proper symmetry with Scala's raw numerics - WDYT?

fthomas:
Yes, makes sense. Any idea why it starts with an upper case `M`? IMHO `PosInt.maxValue` looks much nicer.

howyp:
I'd guess it's [this bit of the style guide](https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods). Which, in my experience, isn't followed that much

build.sbt
```diff
@@ -305,7 +305,7 @@ lazy val compileSettings = Def.settings(
     "-language:higherKinds",
     "-language:implicitConversions",
     "-unchecked",
-    "-Xfatal-warnings",
+//    "-Xfatal-warnings",
```
howyp: I couldn't find a way to allow for the deprecation warnings - is there something better I can do apart from commenting this out?

modules/core/shared/src/main/scala/eu/timepit/refined/api/RefinedTypeOps.scala
```diff
@@ -39,3 +39,10 @@ class RefinedTypeOps[FTP, T](implicit rt: RefinedType.AuxT[FTP, T]) extends Seri
   def unsafeFrom(t: T): FTP =
     rt.unsafeRefine(t)
 }
+object RefinedTypeOps {
+  class Numeric[FTP: Min: Max, T](implicit rt: RefinedType.AuxT[FTP, T])
```
howyp: Do we want this to be a subclass of `RefinedTypeOps`, or a mixin for it? I can see arguments either way. For instance, I doubt we'll ever be able to support `MinValue` and `MaxValue` for all numeric refined types, so maybe this would be better as:

```scala
trait MaxAndMin { this: RefinedTypeOps =>
  ...
}
```

But then I don't think we can ensure that it supports Min/Max by capturing their instances when the companion object is compiled.

modules/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/numeric.scala
```diff
@@ -1,82 +1,75 @@
...
-  def chooseRefinedNum[F[_, _], T: Numeric: Choose, P](min: F[T, P], max: F[T, P])(
-      implicit rt: RefType[F],
-      v: Validate[T, P]
-  ): Gen[F[T, P]] =
-    Gen.chooseNum(rt.unwrap(min), rt.unwrap(max)).filter(v.isValid).map(rt.unsafeWrap)
-
-  ///
-
+trait DeprecatedNumericInstances {
+  @deprecated("Use numericMaxAndMinArbitrary instead", "0.8.6")
```
howyp: I've tried to keep the old arbitrary instances for backwards compat - sadly this only seems to have worked on Scala 2.11. Any good ideas how to make this work?

howyp:
I think this is ready for review!

fthomas:
@howyp Would you use the `Min` and `Max` type classes in user code directly or do you think they are only a implementation detail for the `{Min,Max}Value` methods on companions?

modules/core/shared/src/main/scala/eu/timepit/refined/api/Min.scala
```diff
@@ -0,0 +1,51 @@
+package eu.timepit.refined.api
+
+trait Min[FTP] { def min: FTP }
```
fthomas: I'v used `FTP` in cases where the parameter is always a refined type. Since there are also instances for primitive types (`Int`, `Double`, etc.) I'd go with just `T` here. The same applies to `Max`.

modules/core/shared/src/main/scala/eu/timepit/refined/api/Min.scala
```diff
@@ -0,0 +1,51 @@
...
+  def apply[FTP](implicit ev: Max[FTP]): Max[FTP] = ev
+  def instance[FTP](m: FTP): Max[FTP] = new Max[FTP] { def max: FTP = m }
+}
+
+/**
+ * Auxiliary type class that provides the next smaller or next greater
+ * value for a given argument. This is needed for the `Arbitrary`
+ * instances of numeric refined types.
+ */
+trait Adjacent[T] {
```
fthomas: Could you move this into the `internal` package. I guess we don't want to advertise `Adjacent` as part of our API? Also the Scaladoc needs to be updated. It is not about supporting `Arbitrary` instances any more, right?

modules/core/shared/src/main/scala/eu/timepit/refined/api/Min.scala
howyp: Good point - I'll fix

howyp:
I can imagine using the Min and Max typeclasses from user code when using refined types directly (rather than via a type alias/companion obj)

modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
```diff
@@ -181,3 +181,84 @@ private[refined] trait NumericInference {
   ): Greater[A] ==> Greater[B] =
     Inference(nc.gt(wa.value, nc.fromInt(tb())), s"greaterInferenceWitNat(${wa.value}, ${tb()})")
 }
+
+private[refined] trait NumericMin {
```
fthomas: @howyp What do you think about moving all `Min` / `Max` instances into the `Min` / `Max` companions? That would make them self-contained and would mean that you don't need to import `numeric._` to get instances.

howyp: Generally I much prefer that approach than knowing you have to import certain wildcards. But given you have to import things like `numeric._` for most features in refined, would it be consistent? Because the `Min`/`Max` instance are captured at the companion object creation site, users of `MaxValue` wouldn't need any imports from refined.

It makes me wonder whether it would be a good idea for #386 to look at providing imports which give you everything in refined. At the moment, I think defining a refined type requires two package imports and at least one type import? Or did I miss something?

fthomas: The über-import issue is tracked in #184 and I agree it is a good idea.

The current organization of type class instances grew historically and if I would start from fresh I probably would add the `Validate` and `Inference` instances to the companions of the respective predicates. The same could be done for `Min` and `Max` instances in this setting, although I could also imagine putting them in the `Min` / `Max` companions since these type classes are orthogonal to refined's main functionality and its other type classes. `Min` / `Max` also make only sense for numeric types and therefore are narrower in scope than e.g. `Validate`.

I haven't reached a conclusion on this matter yet. Maybe the consistency argument is good enough to leave it as is.

modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
```diff
@@ -181,3 +182,84 @@ private[refined] trait NumericInference {
...
+  implicit val byteMin: Min[Byte] = Min.instance(Byte.MinValue)
+  implicit val shortMin: Min[Short] = Min.instance(Short.MinValue)
+  implicit val intMin: Min[Int] = Min.instance(Int.MinValue)
+  implicit val longMin: Min[Long] = Min.instance(Long.MinValue)
+  implicit val floatMin: Min[Float] = Min.instance(Float.MinValue)
+  implicit val doubleMin: Min[Double] = Min.instance(Double.MinValue)
+  implicit val charMin: Min[Char] = Min.instance(Char.MinValue)
```
fthomas: I would definitely move these into the `Min` companion since the are not related to the predicates in `numeric`.

modules/core/shared/src/main/scala/eu/timepit/refined/api/RefinedTypeOps.scala
fthomas: I don't get the argument against the subclass approach. Why is a subclass bad if we can't support `MinValue` and `MaxValue` for all numeric refined types? Is it because of the name `Numeric`?

fthomas: Another advantage of using a `trait` here, is that we could add `MinValue` / `MaxValue` functions to companions of refined types that only have a lower or upper bound, e.g. `PosBigInt`:
```scala
trait LowerBounded[T] { 
  def minValue(implicit min: Min[T]): T = min.min
}

trait UpperBounded[T] {
  def maxValue(implicit max: Max[T]): T = max.max
}

trait Bounded extends LowerBounded with UpperBounded

//

type PosInt = Int Refined Positive

object PosInt extends RefinedTypeOps[PosInt, Int] with Bounded[PosInt]

//

type PosBigInt = BigInt Refined Positive

object PosBigInt extends RefinedTypeOps[PosBigInt, BigInt] with LowerBounded[PosBigInt]
```
Of course we could accomplish the same with classes, but that would require a little bit of boilerplate.

fthomas:
@howyp Can we defer the ScalaCheck changes to a later PR? This would allow us to be binary compatible with the 0.8 series and to ignore the fatal-warnings issue for now.

We can deprecate and update the ScalaCheck instances when we are moving to version 0.9.

modules/core/shared/src/main/scala/eu/timepit/refined/api/RefinedTypeOps.scala
howyp: The two big advantages I see of classes for this sort of thing are:
- being able to capture the `Min` and `Max` instances at companion object assembly point, so the user of that object doesn't have to think about those details and can be sure that the `MinValue` and `MaxValue` operations will work
- the `MinValue` and `MaxValue` operations can then be `val`s

Having said that, I'm worried that if we keep things as they currently are in this PR, we will have a class called 'numeric' that doesn't apply to all numeric types :-(. We could rename it to 'Bounded' but then you wouldn't be able to mix it with other such operations that we may come up with for numeric refined companion objects in the future.

So, on balance I reckon we should change this to a trait for the moment, and then add the 'Numeric' class when we can be confident that all predicates are covered. So for now we'd have:

```scala
  trait Bounded[FTP] {
    def MinValue(implicit min: Min[FTP]): FTP = min.min
    def MaxValue(implicit max: Max[FTP]): FTP = max.max
  }
```
And then in the future we could add:
```scala
  class Numeric[FTP: Min: Max, T](implicit rt: RefinedType.AuxT[FTP, T])
      extends RefinedTypeOps[FTP, T]
      with Bounded[FTP] {
    val MinValue: FTP = super.MinValue
    val MaxValue: FTP = super.MaxValue
  }
```

modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
howyp: OK, I think that's enough to convince me to move them into the companion objects. Especially that there is some overlap between the `numeric` and `char` packages in that `Min`/`Max` support both (as well as particular `boolean` predicates).

I'll move them over unless you say otherwise.

howyp:
Happy to pull out the scalacheck changes into another PR. I can't see a 0.9.0 branch, if you can create it then I'll target that. 

We'll still need to have the new `Adjacent` typeclass in core - do you want to leave the currrent one non-deprecated for 0.8.x?

howyp:
I've split the new arbitrary instance into a different PR, and moved the `Min`/`Max` instances into their companion objects. So I think the outstanding issue is if we should have a new trait or class for the companion object.

howyp:
I've come up with a fallback way to support all other predicates via relying on `Validate`. It's less efficient but probably the only way to support `Modulo` and its derivatives. And it has a bit of a performance impact on the `And` instance, but I think that should be ok. 

Unless we can think of a problem with this implementation, I suggest now that we apply the same instances to the `Min` typeclass and then keep `class Numeric extends RefinedTypeOps` rather than turning it into a trait. Sorry for the change of mind :-)

fthomas:
Thanks Howy for splitting this PR.

The new fallback instance does not work in all cases:
```scala
scala> Max[Double Refined Even].max
res3: eu.timepit.refined.api.Refined[Double,eu.timepit.refined.numeric.Even] = 1.7976931348623157E308

scala> Max[Double Refined Odd]
java.lang.OutOfMemoryError: GC overhead limit exceeded
  at java.lang.Double.valueOf(Double.java:519)
  at scala.runtime.BoxesRunTime.boxToDouble(BoxesRunTime.java:79)
  at scala.runtime.java8.JFunction1$mcDD$sp.apply(JFunction1$mcDD$sp.java:12)
  at eu.timepit.refined.internal.Adjacent$$anon$1.nextDown(Adjacent.scala:14)
  at eu.timepit.refined.api.LowPriorityMaxInstances.$anonfun$findValid$1(Max.scala:68)
  at eu.timepit.refined.api.LowPriorityMaxInstances$$Lambda$5000/1279451886.apply(Unknown Source)
  at scala.collection.immutable.Stream$.$anonfun$iterate$1(Stream.scala:1205)
  at scala.collection.immutable.Stream$$$Lambda$5003/1639709136.apply(Unknown Source)
  at scala.collection.immutable.Stream$Cons.tail(Stream.scala:1169)
  at scala.collection.immutable.Stream$Cons.tail(Stream.scala:1159)
  at scala.collection.immutable.Stream.filterImpl(Stream.scala:503)
  at scala.collection.immutable.Stream.filterImpl(Stream.scala:201)
  at scala.collection.TraversableLike.filter(TraversableLike.scala:259)
  at scala.collection.TraversableLike.filter$(TraversableLike.scala:259)
  at scala.collection.AbstractTraversable.filter(Traversable.scala:104)
  at eu.timepit.refined.api.LowPriorityMaxInstances.findValid(Max.scala:68)
  at eu.timepit.refined.api.LowPriorityMaxInstances.findValid$(Max.scala:67)
  at eu.timepit.refined.api.Max$.findValid(Max.scala:13)
  at eu.timepit.refined.api.LowPriorityMaxInstances.validateMax(Max.scala:65)
  at eu.timepit.refined.api.LowPriorityMaxInstances.validateMax$(Max.scala:61)
  at eu.timepit.refined.api.Max$.validateMax(Max.scala:13)
  ... 11 elided
```
Even if we fix this OOME by falling back to a while loop, calculating the largest odd `Double` seems to be impractical. This code
```scala
scala> var d = Double.MaxValue
d: Double = 1.7976931348623157E308

scala> while (d % 2.0 == 0.0) { d = Math.nextDown(d) }
```
ran for over 20 minutes until I aborted it.

howyp:
Yes, it definitely has shortcomings. I'd guess that's OOM just from trying to memoize all of the possible doubles inbetween `Double.MaxValue` and an odd number :-(

Though do we expect users to use the `Modulo` based predicates with non-integer types? I can't think of a case where it's useful. If we can't think of such a case, we could restrict the fallback instance to numbers with an `Integral`.

Having said that - the impl. can definitely be changed to not end up memoizing all of the discarded values whilst calculating, which should improve things.

fthomas:
Users do all kind of crazy things. :-) I'd prefer that `Max[Double Refined Odd]` just does not compile instead of running for a very very long time, so your suggestion with the `Integral` bound sounds good to me. I'd also change the implementation to not use `Stream` since we do not need memoization here.

howyp:
I've switched the fallback algo to use internal mutability for performance, and rebased against the current master. This means that we no longer have issues with `Modulo` and floating-point types, so I've added the fallback implementation for `Min` as well.

fthomas:
Great, thanks @howyp!

PR #384, opened 2017-12-21
Rename cats.validated to cats.syntax

fthomas:
This renames `validated` to `syntax` which would allow us to
use it for future extension methods that are not related to
`Validated`. So this is mostly bikeshedding.

And it also splits the object into an object and a trait.
This split allows for example to define companions which don't
require the `syntax._` import to make the `validate` method
available. The tests show an example of that.

What do you think @howyp?


howyp:
Looks great to me, like the companion object mixin.

PR #392, opened 2017-12-29
Experiment: reorganize type class instances

fthomas:
This is an experiment to find out how moving `Validate` and `Inference` instances to the companions of the predicates affects the library.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -437,7 +437,7 @@ addCommandsAlias(
     "scalafmtTest",
     "scalastyle",
     "test:scalastyle",
-    "mimaReportBinaryIssues",
+    //"mimaReportBinaryIssues",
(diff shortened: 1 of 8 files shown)


fthomas:
The current organization of type class instances will maybe be unsupported in the future, see https://github.com/scala/scala/pull/6226.

howyp:
+1 for making this change. Could it go in 0.9?

Also it looks like you left the Boolean inference typeclasses in a separate trait - is there a reason they have to stay there?

fthomas:
This experiment is not complete, that is the reason why some instances are still in traits.

If this works without breaking (too many) things, I'm fine with doing it in 0.9.

fthomas:
Closing in favor of #438.

PR #393, opened 2017-12-30
Use Min & Max for Scalacheck arbitraries

howyp:
Split from #383. Will rebase once that PR is merged.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -272,7 +272,9 @@ lazy val moduleJvmSettings = Def.settings(
       ProblemFilters.exclude[MissingClassProblem]("eu.timepit.refined.scalacheck.util.OurMath$"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("eu.timepit.refined.StringValidate.*"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("eu.timepit.refined.types.*"),
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("eu.timepit.refined.NumericValidate.*")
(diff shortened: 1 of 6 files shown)


howyp:
Ok, I think I've sorted out the binary compatibility issues in this by leaving the old instances in the same place but marking them as non-implicit. Though I had to add a MIMA exclusion for the new instance, is that an issue? I'm not sure I understand what the "ReversedMissingMethodProblem" is :-(

howyp:
An issue with this change is that because of the fallback instances for `Min` and `Max`, `Arbitrary`s will be resolved for numeric types based on `Modulo` which won't produce valid values. I'll have a think about how to fix this.

fthomas:
I think we can't fix `numericMaxAndMinArbitrary` without filtering for only valid values. This instance needs to work for all predicates and not only those that are defined in the library. Suppose someone has a `IsPrime` predicate and the refined type `type PrimeInt = Int Refined IsPrime`. Valid `Min` and `Max` instances can be defined for `PrimeInt` which means `numericMaxAndMinArbitrary` would be able to give us a `Arbitrary[PrimeInt]` which chooses arbitrary `Int`s between the minimal and maximal prime `Int` and most of these generated values won't satisfy the `IsPrime` predicate.

But filtering only valid values via `Validate` is not a viable option. In the above case ScalaCheck will probably fail with a "discarded too many values" error.
Importing `eu.timepit.refined.scalacheck.numeric._` should only bring implicit `Arbitrary` instances in scope that are not defined with `filter`.

fthomas:
I'm going to close this in favor of #482 which just replaces the usage of `Bounded` with `Min` and `Max`. @howyp if you find other ways to reduce the number of numeric instances, I'm all ears.

PR #402, opened 2018-01-03
Add findNext{Up,Down} to Adjacent

fthomas:
This adds `findNext{Up,Down}` to `Adjacent` so that
* similar code is adjacent :-)
* and `Max.findValid` and `Min.findValid` use a similar implementation

WDYT @howyp?

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/api/Max.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/api/Max.scala
@@ -65,5 +65,5 @@ trait LowPriorityMaxInstances {
     Max.instance(rt.unsafeWrap(findValid(m.max)))
 
   protected def findValid[T, P](from: T)(implicit v: Validate[T, P], a: Adjacent[T]) =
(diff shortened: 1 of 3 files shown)


howyp:
It would be good to put these functions somewhere more logical, but I'm not convinced that they should live in the `Adjacent` typeclass, perhaps because I can't think of a good case where instances would want to override them. I think I'd be ok with them in the companion object for `Adjacent` though.

I've been mulling over how we might implement your `InhabitantsOf` typeclass, now that we have `Max` and `Min` merged, and thinking that it might want to take instances of `Adjacent` that were specific to the refined type. These should be able to be made efficient based on the refinement - for instance `Even` would add `2` every time `nextUp` is called - but would need to have a fallback implementation that is probably based on `findNext{Up,Down}`. So should `findNext{Up,Down}` really be an implementation of `Adjacent` rather than extra methods on it?

fthomas:
Ok, I moved them to the companion and also fixed an overflow errors in the `Numeric` instance.

fthomas:
This PR became stale. Closing for now.

PR #403, opened 2018-01-03
Do not use Numeric#toDouble in Equal's Validate instance

fthomas:
... because of https://github.com/fthomas/refined/issues/396

Using Numeric#fromInt should be safer for this instance.


howyp:
👍

PR #411, opened 2018-01-12
Use numeric type aliases for Min/Max instances

fthomas:
Like `LessEqual[N]` and `GreaterEqual[N]` instead of `Not[Greater[N]]`
and `Not[Less[N]]`. For me the non-negated predicates are easier to
understand.


howyp:
Good idea.

PR #415, opened 2018-01-16
Add Arbitrary instance for NonEmpty Strings

fthomas:


fthomas:
This is not binary compatible with 0.8 and therefore needs to wait until 0.9.

fommil:
you might want to do a `.toString` on the `char` to avoid an implicit. Wartremover and scalafix usually catch that.

PR #423, opened 2018-01-30
Add Arbitrary instance for refinements with a String base type and NonEmpty predicate.

matthedude:
This PR adds an `Arbitrary` instance for refinements with a `String` base type and `NonEmpty` predicate instead of relying on the `any` instance which can cause lots of tests to be discarded.


fthomas:
Thanks @matthedude. There is already an open PR (#415) that adds exactly the same instance. :-)

matthedude:
Ah sorry should have checked! Thanks for all the great work, I'll try and be quicker next time. :)

PR #430, opened 2018-02-01
Compile coreJVM with 2.13.0-M3 on Travis CI

fthomas:


fthomas:
As expected, the build currently fails because of unresolved dependencies:
```
unresolved dependency: org.typelevel#macro-compat_2.13.0-M3;1.1.1: not found
unresolved dependency: com.chuusai#shapeless_2.13.0-M3;2.3.3: not found
unresolved dependency: org.scalamacros#paradise_2.13.0-M3;2.1.1: not found
```

PR #438, opened 2018-02-08
Reorganize type class instances

fthomas:
This is the same as #392 but on top of current master.


fthomas:
With https://github.com/fthomas/refined/pull/438/commits/3c546f6037e10ad27ebe9e6685fcada1d0e8f924 many of the numeric inference tests failed with Scala 2.10. Let's see if they become green again after moving the boolean inference instances.

fthomas:
I've decided to limit this PR only to `Validate` instances because I had trouble moving the `Inference` instances without breaking tests. It is also not clear where put some of the instances that involve more than one predicate. For example, should `implicit def existsNonEmptyInference[P]: Exists[P] ==> NonEmpty` be in the `Forall`, `Empty`, or `Not` companion?

PR #442, opened 2018-02-10
Move Eval and JSONPath instances into companions

fthomas:


fthomas:
I forgot to do this in #438.

PR #447, opened 2018-03-01
automatic typeclass derivation

fommil:
https://gitlab.com/fommil/scalaz-deriving/issues/56

we can also do covariant functors if the provide a `MonadError[?, String]` but I'm encountering some implicit problems... so raising this PR to discuss.


fommil:
I think I need a few SAM type hints to keep the older scala versions happy. Or just not use SAM types.

fommil:
the 2.10 compile error is happening in a file I didn't touch... what's going on?

fthomas:
The first error in the 2.10 build is
```
[info] Compiling 4 Scala sources to /home/travis/build/fthomas/refined/modules/scalaz/js/target/scala-2.10/test-classes ...
[error] /home/travis/build/fthomas/refined/modules/scalaz/shared/src/test/scala/eu/timepit/refined/scalaz/RefTypeSpecScalazMonadError.scala:48:26: diverging implicit expansion for type eu.timepit.refined.scalaz.Decoder[eu.timepit.refined.api.Refined[String,eu.timepit.refined.collection.NonEmpty]]
[error] starting with method refTypeContravariant in package scalaz
[error]     val decoder = Decoder[String Refined NonEmpty]
[error]                          ^
```
which is in a file you've added :-)

fommil:
@fthomas heh, yeah but one of the other files is also failing. I think 2.10's implicit resolution is just not good enough for this derivation. If you can think of a hack, that would be good, otherwise perhaps we should make this 2.11+ only?

fthomas:
I tried shuffling the instances around but that didn't make a difference. I'm fine with making that 2.11+ only.

fommil:
@fthomas green lights!

fthomas:
Looks great, many thanks @fommil!

This is also a great example how to combine type classes via abstractions like `Contravariant` and `MonadError`. I find it fascinating that we can express these relationships in code.

fommil:
@fthomas I'm trying to explore as much of the automatic derivation space as possible for the Typeclass Derivation chapter of my book. It was really good fun figuring this out and supporting refined types turned out to be much more elegant than I'd hoped. Create an iso to a readert and you get free decodings, nice!

PR #449, opened 2018-03-04
Port of #447 for Cats

fthomas:
This is a port of #447 for Cats. One reason for doing this is to see if the 2.10 build fails in a similar way.

/cc @fommil

### DIFF
--- a/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/package.scala
+++ b/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/package.scala
@@ -1,9 +1,11 @@
 package eu.timepit.refined
 
+import _root_.cats.Contravariant
+import _root_.cats.MonadError
 import _root_.cats.Show
 import _root_.cats.implicits._
 import _root_.cats.kernel.{Eq, Order}
-import eu.timepit.refined.api.RefType
+import eu.timepit.refined.api.{RefType, Validate}
 
 package object cats {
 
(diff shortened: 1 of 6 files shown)


fommil:
Damn.

How do you want to do the 2.11+ restriction? I'm not sure how to do that in your codebase.

fthomas:
The first commit in this PR failed on all Scala versions with:
```
[error] /home/travis/build/fthomas/refined/modules/cats/shared/src/test/scala/eu/timepit/refined/cats/RefTypeMonadErrorSpec.scala:78:26: diverging implicit expansion for type eu.timepit.refined.api.RefType[F]
[error] starting with method refTypeMonadError in package cats
[error]     val decoder = Decoder[PosInt]
[error]                          ^
```
Commenting out a seemingly unrelated test in https://github.com/fthomas/refined/pull/449/commits/bdc0b26fd26d247310335267e2206c99695de85d fixed the 2.11+ builds while the 2.10 build still fails with that error.

fthomas:
@fommil The easiest option to make #447 2.11+ only would be to not run the tests with 2.10. In this PR I choose to run the test for `refTypeMonadError` only with 2.12. It would be great if the build would support `scala-2.11+` source directories but this is not the case right now. So until `scala-2.11+` directories are supported or 2.10 support is dropped I'm fine with testing that instance only with 2.12.

fommil:
Ok great, but the addition of the implicit is breaking one of the existing tests in a way I don't understand.

fthomas:
I'm also okay with not running `RefTypeSpecScalazTag` with 2.10 and making it 2.12 only.

fommil:
great! So how do we do that from a practical point of view? I'm not sure how to disable a test.

fthomas:
Moving the test file from refined/modules/scalaz/shared/src/test/scala/eu/timepit/refined/scalaz/ to refined/modules/scalaz/shared/src/test/scala-2.12/eu/timepit/refined/scalaz/ does the trick.

fommil:
Will you cut a release soon with these new things?

fthomas:
I'll want to process some of the open PRs and need to update the release notes before cutting the next release. That will probably take me one or two weeks.

Issue #448, opened 2018-03-02
Creating new refinement type crashes with: exception during macro expansion


umbreak:
I created a [simple project](https://github.com/umbreak/refine-test) which reproduces that error.

When defining my own refined type and using it in the code it fails with the following error at `compile` time:
```
[error].../Const.scala:6:32: exception during macro expansion: 
[error] java.lang.ClassNotFoundException: com.test.refined.Refinements$
[error]         at scala.reflect.internal.util.AbstractFileClassLoader.findClass(AbstractFileClassLoader.scala:64)
[error]         at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[error]         at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[error]         at __wrapper$3$b819929bb64041e38491d57bb6d81ca3.__wrapper$3$b819929bb64041e38491d57bb6d81ca3$.wrapper(<no source file>:6)
[error]         at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]         at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error]         at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]         at java.lang.reflect.Method.invoke(Method.java:498)
[error]         at scala.tools.reflect.ToolBoxFactory$ToolBoxImpl$ToolBoxGlobal.$anonfun$compile$11(ToolBoxFactory.scala:279)
[error]         at scala.tools.reflect.ToolBoxFactory$ToolBoxImpl.eval(ToolBoxFactory.scala:448)
[error]         at scala.reflect.macros.contexts.Evals.eval(Evals.scala:20)
[error]         at scala.reflect.macros.contexts.Evals.eval$(Evals.scala:14)
[error]         at scala.reflect.macros.contexts.Context.eval(Context.scala:6)
[error]         at eu.timepit.refined.macros.MacroUtils.$anonfun$eval$1(MacroUtils.scala:24)
[error]         at scala.Option.getOrElse(Option.scala:121)
[error]         at eu.timepit.refined.macros.MacroUtils.tryN(MacroUtils.scala:28)
[error]         at eu.timepit.refined.macros.MacroUtils.tryN$(MacroUtils.scala:27)
[error]         at eu.timepit.refined.macros.RefineMacro.tryN(RefineMacro.scala:11)
[...]

PR #451, opened 2018-03-08
Added documentation to solve exception during macro expansion errors

umbreak:
Added documentation to solve `java.lang.ClassNotFoundException` when adding a custom predicate inside a project (not REPL).

This addresses the issue #448


fthomas:
LGTM. Many thanks @umbreak!

PR #452, opened 2018-03-10
Add inference rule for `Greater[A] ==> Not[Less[A]]`

zainab-ali:
If a value is greater than `x`, then it also must not be less than `x`.

What do you think of introducing it as an inference rule?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -316,10 +316,10 @@ lazy val moduleJvmSettings = Def.settings(
         "eu.timepit.refined.jsonpath.string.jsonPathValidate"),
       ProblemFilters.exclude[MissingTypesProblem]("eu.timepit.refined.jsonpath.string$*"),
       ProblemFilters.exclude[MissingClassProblem]("eu.timepit.refined.jsonpath.StringValidate"),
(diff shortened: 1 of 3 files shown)


zainab-ali:
The binary compatability check is failing in 2.10 and 2.11 due to the addition of the new method

fthomas:
This LGTM, thanks @zainab-ali! What do you think about using `GreaterEqual[A]` (which is an alias for `Not[Less[A]]`) for the type of that rule? I think `Greater[A] ==> GreaterEqual[A]` is easier to understand. 

If we add this, it would also make sense to add `Less[A] ==> LessEqual[A]`. We can add this in a later PR, no need to do this here.

With regards to the MiMa checks: the next version will be 0.9.0 so it is okay to break binary compatability. Just add MiMa exclusions [here](https://github.com/fthomas/refined/blob/7080ca017f6571723a5885590647f6d5e2d4be42/build.sbt#L271) to ignore those errors.

zainab-ali:
Right, `GreaterEqual` is easier to reason about.  I've thrown `LessEqual` too, since it was straightforward.

Integrating mima and scalafmt into the build like this is a great idea :)  I'll start doing that for my own projects!

PR #453, opened 2018-03-10
Message digests

NeQuissimus:
- Add `HexString` to string types
- Add new category "message digests"
- Add `MD5`, `SHA1`, `SHA224`, `SHA256`, `SHA384`, `SHA512`


NeQuissimus:
@fthomas I don't understand that build error. I import `Refined` at the top...

fthomas:
Wow, this is strange. I've no idea what could be causing this.

modules/core/shared/src/main/scala/eu/timepit/refined/types/digests.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import eu.timepit.refined.W
+import eu.timepit.refined.api.{Refined, RefinedTypeOps}
+import eu.timepit.refined.boolean.And
+import eu.timepit.refined.collection.Size
+import eu.timepit.refined.generic.Equal
+
+/** Module for type representing message digests. */
+object digests {
+  import string.HexStringSpec
```
fthomas: The build error on 2.10 can be fixed by moving this import to the top-level imports.

NeQuissimus: I hadn't had time to look at this, thx :)
How did you figure that out? :D

fthomas: I moved and commented stuff out until it compiled. One of my special powers! ;-)

NeQuissimus:
@fthomas Can you kick Travis once more? Things build with 2.10 locally just fine now...

fthomas:
The build is green now. :+1:

modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
```diff
@@ -17,6 +17,11 @@ object string {
   type TrimmedString = String Refined MatchesRegex[W.`"""^(?!\\s).*(?<!\\s)"""`.T]
 
   object TrimmedString extends RefinedTypeOps[TrimmedString, String]
+
+  /** A `String` representing a hexadecimal number */
+  type HexStringSpec = MatchesRegex[W.`"""[0-9a-f]+"""`.T]
```
fthomas: @NeQuissimus Should we also allow upper case letters `A-F`? What about an optional `0x` prefix? I think that a prefix would be fine for `HexString` but may be inappropriate for the message digests types. Maybe `HexStringSpec` should be private to `digests` to sidestep these potential conflicts?

NeQuissimus: The prefix is not really much more than an artifact to identify a hexadecimal number to humans or when the information of it being a hex number is not available. I think that is exactly the information we are adding with the refined type.
I will make the change to allow for upper-case letters (but not mixed-upper/lower).

modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
```diff
@@ -17,6 +17,11 @@ object string {
   type TrimmedString = String Refined MatchesRegex[W.`"""^(?!\\s).*(?<!\\s)"""`.T]
 
   object TrimmedString extends RefinedTypeOps[TrimmedString, String]
+
+  /** A `String` representing a hexadecimal number */
+  type HexStringSpec = MatchesRegex[W.`"""([0-9a-f]+)|([0-9A-F]+)"""`.T]
```
fthomas: It seems that this does not work with Scala.js (there is a test failure), but fortunately you've [experience with such problems](https://github.com/fthomas/refined/pull/357#issuecomment-359003701). :smiley_cat:

NeQuissimus: Oh god... I shall stop using `coreJVM/test` and dive into regex' in JS once again :D

NeQuissimus:
@fthomas Can you kick the build, some 503 thing just happened on Travis.

PR #455, opened 2018-03-13
Add arbitrary instance for lists with a MaxSize.

matthedude:
I thought I would get the ball rolling for collection arbitrary instances for ScalaCheck. I've added one instance for lists with a `MaxSize`, from here if the structure is OK I can add more as can other people.

### DIFF
--- a/modules/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/all.scala
+++ b/modules/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/all.scala
@@ -7,3 +7,4 @@ object all
     with NumericInstances
     with RefTypeInstances
     with StringInstances
+    with CollectionInstances
--- a/modules/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/collection.scala
+++ b/modules/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/collection.scala
@@ -3,19 +3,21 @@ package eu.timepit.refined.scalacheck
 import eu.timepit.refined.api.{Refined, RefType}
 import eu.timepit.refined.collection.Size
 import org.scalacheck.{Arbitrary, Gen}
+import shapeless.Nat
+import shapeless.ops.nat.ToInt
 
 /** Module that provides `Arbitrary` instances for collection predicates. */
 object collection extends CollectionInstances
 
 trait CollectionInstances {
 
-  implicit def listSizeArbitrary[T: Arbitrary, F[_, _]: RefType, P](
-      implicit arb: Arbitrary[Int Refined P]
-  ): Arbitrary[F[List[T], Size[P]]] =
-    arbitraryRefType[F, List[T], Size[P]](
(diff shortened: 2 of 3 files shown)


fthomas:
I wanted to have `Arbitrary` instances for collection predicates for a long time. So thanks for getting the ball rolling. The structure is certainly fine and I'd be okay with merging this right away. But if you'd like to iterate on that, I've a suggestion how to make that instance more generic. I think we can change the type of that instance to `Arbitrary[F[List[T], Size[P]]]` by using the numeric `Arbitrary` instances. The reason for this is that `MaxSize[N]` is an alias for `Size[LessEqual[N]]` and our numeric instances already provide e.g an `Arbitrary[Int Refined LessEqual[N]]`.
If I'm not mistaken, something like this should work:
```scala
implicit def listSizeArbitrary[T: Arbitrary, F[_, _]: RefType, P](
      implicit arb: Arbitrary[Int Refined P]
  ): Arbitrary[F[List[T], Size[P]]] = ...
```
`arb` should give us the correct `Int` for `Gen.listOfN` such that its length satisfies the predicate `P`. That instance should now work for `MaxSize[N]` and `MinSize[N]` alike.

Let me know if you'd like to give that a try.

matthedude:
Ah I get your point about using the numeric instances, makes sense especially as that was where I found the `Nat` technique. Great I'll give it a go thanks for the info 👍

matthedude:
This worked exactly as you thought it would. However I'm struggling a bit with testing `List[T] Refined MinSize[N]` as it generates huge lists that bork everything. As you can see I've now introduced an upper bound for the size of a `List` of 999 but I'm wondering if there is a better way of doing this, I'm not hugely experienced with ScalaCheck so I'm probably missing something obvious!

fthomas:
I'd be okay to go with 1895b52abbe27c7f273e4368901f101b75389232 and just not test that instance with `MinSize[N]` but with e.g. `Interval.Closed[_3, _13]` instead. The current upper bound `N + 999` for the predicate `Size[P[N]]` might not satisfy `P` so I'm afraid we can't do this.

I think what we could do is to skew this instance towards shorter lists. I imagine something like this: if `arb: Arbitrary[Int Refined P]` gives us an `Int` that is bigger than some bound (999), we could ask it again to give us e.g. 5 more `Int`s from which we use the smallest one. I'm not sure if this works out in practice and I'm fine with doing this later as an optimization.

I've to think about adding a lower bound of 0 to `Size` a bit more. For now I'd filter `arb` such that it does not produce negative `Int`s.

Oh, and if you think this generalisation has opened a can of worms, we could also go back to you first commit and merge that right away. :-)

matthedude:
Thanks @fthomas really helpful points again. I'm going to look at it later and try out your suggestions and hopefully keep it generalised and useful. If I fail I'll let you know and we can merge one of the earlier commits so cleverer people than me can come along and help improve it :)

matthedude:
I've put it back to the generalised `Size[P]` version with no limits and then only tests for the suggested "safe" predicates. I had a few problems trying to use `Nat` with the `Interval` but then noticed this probably doesn't work yet as there are no `numeric` tests for `Interval` with a `Nat`.

PR #456, opened 2018-03-17
Added support to generate a Typeable for a refined type

umbreak:
Added a module for shapeless support of `Typeable`


umbreak:
It would be great if someone can take a look at it, since the I don't see an obvious error for this to fail. Only the build in [2.10.7](https://travis-ci.org/fthomas/refined/jobs/355739103) is failing without any message.

Issue #437, opened 2018-02-07
Add FiniteString?


fthomas:
https://medium.com/anyjunk/fstring-string-lengths-at-compile-time-81f811750e9e is an interesting article about `String`s with a specified maximum length in the type. In refined we can do something similar with the `MaxSize` predicate: `type FiniteString[N] = String Refined MaxSize[N]`. Maybe this refined type with some convenience functions would make a nice addition to the library?

PR #471, opened 2018-04-10
Arbitrary instances for Vector/String with Size[P]

fthomas:
This builds up on the work @matthedude did in #455 to provide `Arbitrary` instances for `Vector` and `String` with any `Size[P]` predicate.

The instance for `String` will become handy if we add something like `FiniteString[N]` (#437) to the library.


matthedude:
This is really cool great work 👍

PR #474, opened 2018-04-11
Drop support for Scala 2.10

fthomas:


fthomas:
From the list of external modules, these are still available for 2.10:
- argonaut-refined
- atto-refined
- ciris-refined
- monocle-refined
- scala-xml-names

Dropping 2.10 support would make it harder for those downstream projects to update refined, so I'll defer this for now.

fthomas:
Ciris dropped 2.10 support (https://github.com/vlovgr/ciris/pull/161) and Monocle is also considering dropping 2.10 support (https://github.com/julien-truffaut/Monocle/issues/575#issuecomment-399428009). I think we can do this now after the 0.9.1 release.

PR #487, opened 2018-04-26
Add a TrimmedString.trim method

ceedubs:
This is somewhat similar to `FiniteStringOps.truncate` in that it's
guaranteed to turn any `String` into a `TrimmedString`.

There are some funny characters that get removed by `String.trim` but
don't match on `\s` in regular expressions. I've added an example to
demonstrate that these characters can lead to a `String` being "trimmed"
a bit more than would technically be necessary to match the
`TrimmedString` regex.

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
@@ -43,8 +43,16 @@ object string {
 
   object NonEmptyString extends RefinedTypeOps[NonEmptyString, String]
 
-  /** A `String` that contains no leading or trailing whitespace. */
-  type TrimmedString = String Refined MatchesRegex[W.`"""^(?!\\s).*(?<!\\s)"""`.T]
+  // scalastyle:off no.whitespace.after.left.bracket
+  /**
+   * A `String` that contains no leading or trailing whitespace.
+   *
+   * Note that a line separator (`\u2028') is not considered whitespace for the
+   * purposes of trimming.
+   */
+  type TrimmedString = String Refined MatchesRegex[
+    W.`"""^(?![\\x{0000}-\\x{0020}])(?s:.*)(?<![\\x{0000}-\\x{0020}])"""`.T]
+  // scalastyle:on no.whitespace.after.left.bracket
 
   object TrimmedString extends RefinedTypeOps[TrimmedString, String] {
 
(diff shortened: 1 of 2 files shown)


ceedubs:
Oops I have a broken test :(

ceedubs:
Strings are hard. I'm not sure what to do here.

fthomas:
So [`String#trim`](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#trim()) removes all leading and trailing characters with a code less than or equal to `\u0020`. I think we can replace `\s` in the regex of `TrimmedString` to `[\x{0000}-\x{0020}]` to mimic the semantics of `String#trim`. I'm in favor of doing this although it changes the semantics of the refined type. But breaking the expectation that every `s.trim` is a valid `TrimmedString` is much worse IMO.

ceedubs:
@fthomas thanks for the guidance!

I pushed 9a9a877a22b4ad8d49ef346712551c00ebe5c1d9 which hopefully addresses the issue. I used your advice but also had to make an additional slight change. See the commit message for more details.

ceedubs:
Okay I've upgraded to a `NullPointerException` as my build failure. Hmmm.

ceedubs:
I'm not sure what the build failure is about. Any ideas? Also locally I have a change to add a scalacheck generator for trimmed strings -- should I add that here or would you prefer that as a separate PR afterwards?

fthomas:
The test passes on the JVM but fails with JS. Oh no. :-( This is not the first time that we have difficulties writing regexes that should work on the JVM and with JS (see https://github.com/fthomas/refined/pull/453#discussion_r181738795 and https://github.com/fthomas/refined/pull/357#issuecomment-359003701). I've no idea if the current regex can be changed to work on both platforms. I'm at a point where I would suggest adding a `Trimmed` predicate that is implemented with `String#trim` and redefine `TrimmedString` as `String Refined Trimmed`. I bet this would be easier than changing the regex to work on both platforms and it would probably be more robust because it doesn't depend on the peculiarities of the underlying regex engine.

I'm sorry this is not a straightforward change as it should have been.

fthomas:
Instead of adding `Trimmed` now you could also make the failing test JVM-only, and we add that predicate later. I could imagine that the original definition of `TrimmedString` already does not work with JS which is possible because there are currently no tests. So merging this PR now would not make anything worse.

WRT the ScalaCheck instances I'm fine with both.

ceedubs:
Thanks again for the guidance @fthomas! I'm inclined to agree with your suggestion of adding a `Trimmed` predicate. I'll plan on making that change. I'll go ahead and close this PR out and start a fresh new one.

PR #491, opened 2018-04-27
Experiment: replace Witness with proper literal types

fthomas:
This PR replaces `shapeless.Witness` with proper literal singleton types that will be available with Scala 2.13.

:sparkles: This is a glimpse into a better world. :sparkles:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,15 +26,15 @@ notifications:
     on_start: false
 matrix:
   include:
-  - scala: 2.10.7
-  - scala: 2.11.12 # Remember to update this in build.sbt, too.
-    sudo: required
-    before_install:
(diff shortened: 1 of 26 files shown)


fthomas:
I won't merge this as is because it only works with 2.13. I'd like to find out if we can write scalafix rules that replace `ValueOf[T]` with `Witness.Aux[T]` and proper literal types `$x` with ``W.$x.T`` when building with Scala versions prior to 2.13. This way we could live in the better world while still be able to cross-build to older Scala versions.

fthomas:
We're now duplicating source files to be able to use proper literal types with 2.13 and Dotty, see https://github.com/fthomas/refined/pull/828.

PR #493, opened 2018-04-29
Improve specs

rider-yi:
Added some tests for refined types


kusamakura:
I think as in https://github.com/fthomas/refined/pull/497#issuecomment-389100122, just test the `from` method. That should cut down on most of the copy paste work :)

rider-yi:
Done

fthomas:
LGTM, many thanks @rider-yi!

PR #497, opened 2018-05-01
Add numeric types

sh0hei:
- Byte
  - PosByte
  - NonNegByte
  - NegByte
  - NonPosByte

- Short
  - PosShort
  - NonNegShort
  - NegShort
  - NonPosShort


fthomas:
@sh0hei MiMa reported that this change is not binary compatible. Moving the forwarders in `NumericTypes` to a new `NumericTypesBinCompat1` should make this binary compatible. Something similar was done in https://github.com/fthomas/refined/pull/505/commits/2fc946f27450b5edab6a165b57b3023cd373fcc5.

Could also please only test the `from` function on each type? I don't think it is necessary to test `unsafeFrom` or `unapply` on each type since these are implemented in terms of `from` and we've already checked elsewhere that `unsafeFrom` and `unapply` are consistent with `from`.

PR #504, opened 2018-05-10
Replace TrimmedString regex with Trimmed predicate type

ceedubs:
This also adds `TrimmedString.trim` and is a replacement for #487 based
on the discussion there.

While we are at it I threw in an `Arbitrary` instance for `TrimedString`.

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -42,7 +42,7 @@
  </check>
  <check level="error" class="org.scalastyle.scalariform.NumberOfTypesChecker" enabled="true">
   <parameters>
-   <parameter name="maxTypes"><![CDATA[40]]></parameter>
+   <parameter name="maxTypes"><![CDATA[45]]></parameter>
   </parameters>
  </check>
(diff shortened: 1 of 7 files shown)


scalastyle-config.xml
```diff
@@ -42,7 +42,7 @@
  </check>
  <check level="error" class="org.scalastyle.scalariform.NumberOfTypesChecker" enabled="true">
   <parameters>
-   <parameter name="maxTypes"><![CDATA[40]]></parameter>
+   <parameter name="maxTypes"><![CDATA[45]]></parameter>
```
ceedubs: Adding `Trimmed` bumped us over the limit here. 41 seemed a bit too arbitrary so I bumped this to 45. I will not attempt to defend this solution.

ceedubs:
Thanks @fthomas. I've hopefully fixed the binary compatibility.

fthomas:
Looks great, thanks again @ceedubs!

PR #522, opened 2018-05-29
Add ISO8601 dates

NeQuissimus:
Add ISO-8601 formatted Strings

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -315,7 +315,11 @@ lazy val moduleJvmSettings = Def.settings(
   mimaBinaryIssueFilters ++= {
     import com.typesafe.tools.mima.core._
     Seq(
-      )
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
+        "eu.timepit.refined.types.TimeTypes.ISO8601"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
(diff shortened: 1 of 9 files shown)


NeQuissimus:
@fthomas Help :)

It seems that the line with the regex either makes `scalafmt` upset because the line is too long or `scalastyle` is upset because `scalafmt` wants to break the line after the `[`.

I tried a couple of other ways of breaking the line and none seem to work...

NeQuissimus:
Removing the scalacheck rule seems to have done the trick

fthomas:
I've a few comments and suggestions:
* https://en.wikipedia.org/wiki/ISO_8601 talks about *basic* and *extended* formats but the regex requires the *extended* format, right? It also seems that the standard allows to drop seconds and minutes which does not seem to be supported by the regex. So if the type does not fit all ISO 8601 formatted dates, a more specific name than `ISO8601` would be good.
* I'd prefer to update the tut docs in a separate PR
* To make this change binary compatible we need to add a new trait `TimeTypesBinCompat1` similar to this change https://github.com/fthomas/refined/commit/2fc946f27450b5edab6a165b57b3023cd373fcc5. We can then remove the MiMa exceptions.
* One or more tests would be nice.

PR #530, opened 2018-06-19
Use sbt-release-early

### LINKED ISSUE #43: Let Travis publish snapshots to Sonatype

fthomas:
closes #43


fthomas:
One downside of the current approach is that the CI now takes between 28 and 48 minutes because there are so many modules that need to be released.

fthomas:
Closing in favor of #603.

PR #539, opened 2018-07-06
Add syntax for Cats NonEmptyList

cb372:
Adds a `refinedSize` extension method that returns the size of a NonEmptyList as a positive int.

Based on [this Twitter thread](https://twitter.com/LukaJacobowitz/status/1015216561152577536), at least 2 people in the world want this 😄


fthomas:
This is great, thanks @cb372!

We could also have the same for the other `NonEmpty` types in Cats and also for the standard library collections to return a `NonNegInt`. Maybe there are more methods where we could provide a more precise return type.

Issue #450, opened 2018-03-08
NullPointerException raised when attempting conversion on null string


umbreak:
I have a type defined as:
```type Some = String Refined MatchesRegex[W.`"[a-zA-Z_][a-zA-Z0-9-_.]*"`.T]```

and then I try to get this type back from a string:
```
RefType.applyRef[Some](string)
```

The result is the following exception on runtime:
```
java.lang.NullPointerException:
at eu.timepit.refined.StringValidate.$anonfun$matchesRegexValidate$1(string.scala:129)
at eu.timepit.refined.StringValidate.$anonfun$matchesRegexValidate$1$adapted(string.scala:129)
at eu.timepit.refined.api.Validate$$anon$3.validate(Validate.scala:79)
at eu.timepit.refined.internal.RefinePartiallyApplied.apply(RefinePartiallyApplied.scala:15)
at eu.timepit.refined.internal.ApplyRefPartiallyApplied.apply(ApplyRefPartiallyApplied.scala:19)
...
```

Since applyRef is returning a `Either[String, Some]`, shouldn't that type of error (the string being null) return a `Left(Not allowed empty string)`?

**Comment by jan0sch** [comment]
Hmm... The same also happens when using `NonEmpty`.

```
import eu.timepit.refined._
import eu.timepit.refined.api._
import eu.timepit.refined.boolean._
import eu.timepit.refined.collection._
import eu.timepit.refined.numeric._
import eu.timepit.refined.string._

@ type SomeString = String Refined MatchesRegex[W.`"[a-zA-Z_][a-zA-Z0-9-_.]*"`.T] 
defined type SomeString

@ type SomeNonEmptyString = String Refined And[NonEmpty, MatchesRegex[W.`"[a-zA-Z_][a-zA-Z0-9-_.]*"`.T]] 
defined type SomeNonEmptyString

@ val n: String = null 
n: String = null

@ RefType.applyRef[SomeString](n) 
java.lang.NullPointerException
  eu.timepit.refined.StringValidate.$anonfun$matchesRegexValidate$1(string.scala:129)
  eu.timepit.refined.StringValidate.$anonfun$matchesRegexValidate$1$adapted(string.scala:129)
  eu.timepit.refined.api.Validate$$anon$3.validate(Validate.scala:79)
  eu.timepit.refined.internal.RefinePartiallyApplied.apply(RefinePartiallyApplied.scala:15)
[...]

PR #548, opened 2018-07-19
Adding NonNull type for validation of potentially null values.

ehoner:
This is a duplicate of fthomas/refined/issues/450, which is a request to provide safe `null` handling.

There is a discussion on the previous PR that I would like to address. I agree with @jan0sch [comment](https://github.com/fthomas/refined/issues/450#issuecomment-375655000) that it is the responsibility of the developer to handle `null`, but it seems reasonable that the library should afford the developer a mechanism to do so. I believe this patch would not cause any breaking changes and the maintenance overhead should be minimal.

### DIFF
--- a/modules/core/shared/src/test/scala/eu/timepit/refined/ImpureValidateSpec.scala
+++ b/modules/core/shared/src/test/scala/eu/timepit/refined/ImpureValidateSpec.scala
@@ -19,8 +19,6 @@ class ImpureValidateSpec extends Properties("ImpureValidate") {
   }
 
   property("NonNull.showResult - String - invalid") = secure {
-    import eu.timepit.refined.collection.NonEmpty
-
     val string: String = null
     showResult[NonNull[NonEmpty]](string) ?= "Predicate failed: source is null."
   }
--- a/modules/core/shared/src/test/scala/eu/timepit/refined/ImpureValidateSpec.scala
+++ b/modules/core/shared/src/test/scala/eu/timepit/refined/ImpureValidateSpec.scala
@@ -1,0 +1,63 @@
+package eu.timepit.refined
+
+
+import eu.timepit.refined.TestUtils._
+import eu.timepit.refined.boolean._
+import eu.timepit.refined.collection._
+import eu.timepit.refined.impure.NonNull
+import eu.timepit.refined.numeric._
+import eu.timepit.refined.string._
+import org.scalacheck.Prop._
+import org.scalacheck.Properties
+
+class ImpureValidateSpec extends Properties("ImpureValidate") {
+
+  property("NonNull safe for String") = secure {
+    import eu.timepit.refined.collection.NonEmpty
+
+    val string: String = null
+    notValid[NonNull[NonEmpty]](string)
+  }
+
+  property("NonNull.showResult - String - invalid") = secure {
+    import eu.timepit.refined.collection.NonEmpty
+
+    val string: String = null
+    showResult[NonNull[NonEmpty]](string) ?= "Predicate failed: source is null."
+  }
+
+  property("NonNull safe for List") = secure {
+    val list: List[Int] = null
+    notValid[NonNull[NonEmpty]](list)
+  }
+
+  property("NonNull valid String") = forAll { string: String =>
+    string.nonEmpty ==>
+      isValid[NonNull[NonEmpty]](string)
+  }
+
+  property("NonNull.showResult - String - left valid") = secure {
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/test/scala/eu/timepit/refined/ImpureValidateSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+import eu.timepit.refined.impure.NonNull
+import eu.timepit.refined.numeric._
+import eu.timepit.refined.string._
+import org.scalacheck.Prop._
+import org.scalacheck.Properties
+
+class ImpureValidateSpec extends Properties("ImpureValidate") {
+
+  property("NonNull safe for String") = secure {
+    import eu.timepit.refined.collection.NonEmpty
```
Michaelt293: I believe `NonEmpty` is already being imported on line 5 -`import eu.timepit.refined.collection._`?

ehoner: You are correct.

jan0sch:
Very nice! Does this work for all types which could be `null` or are there limits to it?

ehoner:
@jan0sch 

I am not sure that I can say *all*, but the tests verify basic types `String` and `List` are safe when  `null`. With an additional test for `Positive` to show compatibility with the library - since `int` types cannot be null. There is also an example, with `URL`, where the `URL` type can be refined with the library (even if the value is `null`), but a valid `Refined[URL]` instance may return a `null` with `getUserInfo` and this PR can handle that safely. 

The goal was to make something specific to `null` handling that was compatible with the rest of the library. FWIW, all the tests that were immediately obvious to me have been included. I stumbled into this when I encountered the problem myself. Another demonstration based on the [comment](https://github.com/fthomas/refined/issues/450#issuecomment-387854403) in another PR, where the `Not` case was previously unsafe.

```scala
scala> val s: String = null
s: String = null

scala> refineV[Regex](s)
res3: Either[String,eu.timepit.refined.api.Refined[String,eu.timepit.refined.predicates.all.Regex]] = Left(Regex predicate failed: null)

scala> refineV[NonNull[Not[Regex]]](s)
res4: Either[String,eu.timepit.refined.api.Refined[String,eu.timepit.refined.predicates.all.NonNull[eu.timepit.refined.predicates.all.Not[eu.timepit.refined.predicates.all.Regex]]]] = Left(Predicate failed: source is null.)
```

ehoner:
@fthomas do you have any thoughts?

fthomas:
Sorry for the late reply, @ehoner.

I'm still wondering what the advantage of this predicate would be over composing `refineV` with a function that handles `null`s:
```scala
scala> def nonnull[T](t: T): Either[String, T] = if (t == null) Left("parameter is null") else Right(t)

scala> nonnull(null: String).flatMap(refineV[NonEmpty](_))
res1: scala.util.Either[String,eu.timepit.refined.api.Refined[String,eu.timepit.refined.predicates.all.NonEmpty]] = Left(parameter is null)

scala> nonnull("").flatMap(refineV[NonEmpty](_))
res2: scala.util.Either[String,eu.timepit.refined.api.Refined[String,eu.timepit.refined.predicates.all.NonEmpty]] = Left(Predicate isEmpty() did not fail.)

scala> nonnull("Hello!").flatMap(refineV[NonEmpty](_))
res3: scala.util.Either[String,eu.timepit.refined.api.Refined[String,eu.timepit.refined.predicates.all.NonEmpty]] = Right(Hello!)
```

ehoner:
@fthomas 

I am not sure that I can definitively answer your question: benefits of function `nonnull` vs type `NonNull`. I can only offer that a type might better identify what refinement has taken place.

From my own experience, I created the PR when I first started using the library. I encountered the `null` problem immediately and my expectation was that the library would provide some mechanism for safely handling this. I didn't find any documentation/examples specific to `null`, but I did find a previous [issue](/issues/450). So I added what I thought was a way to handle `null` that is consistent with `refined`. This addition might be overkill, in fact I can't think of an example why the `NonNull` type would be strictly "necessary".

Baring acceptance of this PR, it might be valuable, for one of the examples to use the function above to help newcomers to the library. Or even add the function to the library to newcomers or provide a "refined" way to handle `null`. 

PS. Thanks for the library.

ehoner:
I believe this is unnecessary and not a valuable addition. I should have closed this long ago, I left it open to think about examples where this would be useful and was not able to identify any. Then forgot about the PR, sorry for letting this linger. 

Thanks again for your help @fthomas and the library.

PR #567, opened 2018-09-13
Numeric refined types for `Byte`, `Short`, `BigInt` and `BigDecimal`.

Michaelt293:
Added numeric refined types for `Byte`, `Short`, `BigInt` and `BigDecimal`.


fthomas:
Thanks for the PR, @Michaelt293! To make this change binary compatible forwarders of all new types must be added to `NumericTypesBinCompat1` and it cannot be mixed into `NumericTypes`. In `types/all.scala` you then need a new trait:
```scala
trait AllTypesBinCompat1 extends NumericTypesBinCompat1
```
which is then mixed into the `all`object. We have a similar situation here: https://github.com/fthomas/refined/blob/master/modules/core/shared/src/main/scala/eu/timepit/refined/predicates/all.scala

Michaelt293:
Thanks for your help @fthomas. Was very helpful :)

PR #573, opened 2018-10-05
Add cross-build for scalacheck 1.13; fixes #572

### LINKED ISSUE #572: refined-scalacheck binary compatibility
Since 0.9.1 the `refined-scalacheck` module depends on scalacheck 1.14, which is [not binary compatible with 1.13 series](https://github.com/rickynils/scalacheck/blob/1.14.0/RELEASE.markdown#whats-new-in-scalacheck-1140).

Although this isn’t a problem for refined itself, depending on instances from `refined-scalacheck` can cause problems when used with something like `cats-laws` or `circe-testkit` since [cats still depends on 1.13](https://github.com/typelevel/cats/issues/2449).

Might it be possible to cross-build `refined-scalacheck` for both 1.13 and 1.14?

liff:
Produces an additional `refined-scalacheck_1.13` module that depends on scalacheck 1.13.


fthomas:
@liff MiMa should stop complaining when you add `refined-scalacheck_1.13` to `unreleasedModules` in `latestVersion.sbt`.

liff:
Yeah, sorry. I sometimes forget to check the build results.

PR #589, opened 2018-11-09
Add syntax module

NeQuissimus:
I could have used this a lot today, so I am finally getting around to contributing a syntax module :)

Basically, these return refined types for a few methods on `NonEmptyString`.
Also, I added a string interpolator that creates `NonEmptyString`.

Anybody is welcome to add to this, I may provide more convenience syntax as I need it.

Notes: I know virtually nothing about Scala.js or Scala.Native. I created the cross-build targets but am unsure whether they should exist...

Edit: Removed Scaja.js and Scala.Native. Looks like the string interpolator does not work there

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,6 +12,7 @@ val gitPubUrl = s"https://github.com/$gitHubOwner/$projectName.git"
 val gitDevUrl = s"git@github.com:$gitHubOwner/$projectName.git"
 
 val catsVersion = "1.4.0"
+val contextualVersion = "1.1.0"
 val jsonpathVersion = "2.4.0"
 val macroParadiseVersion = "2.1.1"
 val pureconfigVersion = "0.10.0"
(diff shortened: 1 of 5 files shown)


fthomas:
Thanks for this, @NeQuissimus! I'd vote to move the NonEmptyStringOps into the core module since it doesn't need new dependencies. And instead of refined-syntax I'd call the new module refined-contextual.

I've also worked on extension methods for `NonEmptyString` here https://github.com/fthomas/refined/compare/topic/NonEmptyString.Syntax but hadn't time to polish it. The methods there are available without any additional import because `NonEmptyString.Syntax` is mixed into the implicit scope of `NonEmptyString`. Not sure if this is the way to go, but it was nice to see that it is possible. Maybe you've any thoughts about that.

NeQuissimus:
I think keeping the implicit conversions behind an import will keep the compile times down for users. I would rather not expose them automatically.

What if we just create a syntax package in the core module?

We can then turn the string interpolation part into `refined-contextual`.

Issue #256, opened 2017-01-27
typechecker.Conexts not found when upgrading from 0.6.2 -> 0.7.0


ngbinh:
Not sure if anyone like me but I get this error when upgrading `refined` from `0.6.2` to `0.7.0`.

```scala
[error] Class scala.tools.nsc.typechecker.Contexts not found - continuing with a stub.
[error 01:39:37:AM] Class scala.tools.nsc.typechecker.Contexts not found - continuing with a stub.
[error] Class scala.tools.nsc.typechecker.Contexts not found - continuing with a stub.
[error 01:39:37:AM] Class scala.tools.nsc.typechecker.Contexts not found - continuing with a stub.
[error] two errors found
```

fthomas:
Oh dear! This is probably caused by #251. Are you using the `Eval` predicate?

fthomas:
I can reproduce this in one of my projects. The strange thing is that it only happens on the first `compile` after starting sbt. Compilation succeeds without errors or warnings on subsequent `compile`s. This is independent of the Scala version (tested with 2.11.8 and 2.12.1) - they only differ in the error message. I get 
```
[error] Class scala.tools.nsc.Global not found - continuing with a stub.
```
on 2.11.8.

ngbinh:
We are using a lot of `Regex` but no `Eval` I believe. It happens on scala `2.12.1` for us

ngbinh:
...and it happens to us consistently

fthomas:
I currently have no idea how to fix this other than reverting #251. @clhodapp, do you have any ideas?

clhodapp:
Yikes! Let's try to gather a bit more detail about what everyone's dependency topologies look like and/or get an open source reproduction.

fthomas:
Here is a reproduction: https://github.com/fthomas/scala-web-stack/pull/4

clhodapp:
Currently, I am thinking that the problem is the use of `eval` in https://github.com/fthomas/refined/blob/master/modules/core/shared/src/main/scala/eu/timepit/refined/macros/RefineMacro.scala. I think that using `Context#eval` in a macro actually creates a compile-time dependency on scala-compiler.

[...]

PR #592, opened 2018-11-29
remove scalac dependency

yanns:
I'm not sure if we need the scalac dependency in `core`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -88,6 +88,9 @@ lazy val benchmark = project
   .dependsOn(coreJVM)
   .enablePlugins(JmhPlugin)
   .settings(noPublishSettings)
+  .settings(
+    libraryDependencies += scalaOrganization.value % "scala-compiler" % scalaVersion.value
+  )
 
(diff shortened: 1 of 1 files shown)


fthomas:
The 2.12 build failed with
```
[error] /home/travis/build/fthomas/refined/modules/core/jvm/target/scala-2.12/src_managed/test/eu/timepit/refined/autoDoctest.scala:22:43: Class scala.tools.nsc.typechecker.Contexts not found - continuing with a stub.
[error]     val x: Int Refined Greater[W.`5`.T] = 100
[error]                                           ^
```
We've seen this error before in #256 which was also caused by removing the scala-compiler dependency in version 0.7.0. We brought it back in 0.8.0.

The dependency is needed for refined's macros. If we want to remove that dependency, we have to move the macros into a new `refined-macros` module. Moving macros into its own module would make it easier to cross-compile refined with dotty, so I'd be in favour of such split.

fthomas:
I've opened an issue for moving the macros to a new module and hence closing this PR.

PR #596, opened 2018-12-12
Added `NotNaN` predicate

dcastro:
As previously discussed in the gitter channel.

There is some duplication for `Float` and `Double`, the only types with a NaN value. I've considered adding a `HasNaN[A]` typeclass to reduce duplication, but ultimately thought the complexity was unwarranted in this case. If the maintainers think otherwise, I'd happy to make this change though.

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
@@ -41,6 +41,9 @@ object numeric extends NumericInference {
   /** Predicate that checks if an integral value modulo `N` is `O`. */
   final case class Modulo[N, O](n: N, o: O)
 
+  /** Predicate that checks if a floating-point number value is not NaN. */
+  case class NotNaN()
+
   /** Predicate that checks if a numeric value is less than or equal to `N`. */
   type LessEqual[N] = Not[Greater[N]]
 
@@ -116,6 +119,14 @@ object numeric extends NumericInference {
         t => s"($t % ${wn.snd} == ${wo.snd})",
         Modulo(wn.fst, wo.fst)
       )
+  }
+
+  object NotNaN {
+    implicit def floatNotNaNValidate: Validate.Plain[Float, NotNaN] = fromIsNaN(_.isNaN)
+    implicit def doubleNotNaNValidate: Validate.Plain[Double, NotNaN] = fromIsNaN(_.isNaN)
+
+    def fromIsNaN[A](isNaN: A => Boolean): Validate.Plain[A, NotNaN] =
+      Validate.fromPredicate(x => !isNaN(x), x => s"$x != NaN", NotNaN())
   }
 }
 
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
@@ -42,7 +42,7 @@ object numeric extends NumericInference {
   final case class Modulo[N, O](n: N, o: O)
 
   /** Predicate that checks if a floating-point number value is not NaN. */
-  case class NotNaN()
+  final case class NonNaN()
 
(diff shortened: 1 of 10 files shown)


modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
```diff
@@ -41,6 +41,9 @@ object numeric extends NumericInference {
   /** Predicate that checks if an integral value modulo `N` is `O`. */
   final case class Modulo[N, O](n: N, o: O)
 
+  /** Predicate that checks if a floating-point number value is not NaN. */
+  case class NotNaN()
```
fthomas: This should be `final`.

fthomas: Other negated predicates use `Non` as prefix, like `NonEmpty`, `NonPositive`, `NonNegative`, or `NonDivisible`. I think we should do it here too and rename it to `NonNaN`. It [seems](https://en.wiktionary.org/wiki/non-) to be a common rule to prefix with non- to negate the meaning of a word.

modules/core/shared/src/main/scala/eu/timepit/refined/types/numeric.scala
```diff
@@ -165,6 +165,16 @@ object numeric {
   type NonPosBigDecimal = BigDecimal Refined NonPositive
 
   object NonPosBigDecimal extends RefinedTypeOps[NonPosBigDecimal, BigDecimal]
+
+  /** A `Float` that is not NaN. */
+  type FloatNotNaN = Float Refined NotNaN
```
fthomas: Since we have `NonNegInt` or `PosDouble`, this should be `NonNaNFloat`.

modules/scalacheck/shared/src/main/scala/eu/timepit/refined/scalacheck/numeric.scala
```diff
@@ -111,3 +111,15 @@ trait NumericInstances {
   ): Arbitrary[F[T, P]] =
     arbitraryRefType(Gen.chooseNum(min, max))
 }
+
+trait NumericInstancesBinCompat1 {
+  implicit def floatNotNaNArbitrary[F[_, _]: RefType](
+      implicit arb: Arbitrary[Float]
+  ): Arbitrary[F[Float, NotNaN]] =
+    arbitraryRefType(arb.arbitrary.suchThat(x => !x.isNaN))
```
fthomas: Minor nitpick: Using `suchThat` could lead in theory to "discarded too many values" errors. I don't think this will ever be a problem in this case but we could eliminate this possibility altogether by mapping `NaN` values to `0.0`.
```suggestion
    arbitraryRefType(arb.arbitrary.map(x => if (x.isNaN) 0.0F else x))
```

modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala
```diff
@@ -117,6 +120,14 @@ object numeric extends NumericInference {
         Modulo(wn.fst, wo.fst)
       )
   }
+
+  object NotNaN {
+    implicit def floatNotNaNValidate: Validate.Plain[Float, NotNaN] = fromIsNaN(_.isNaN)
+    implicit def doubleNotNaNValidate: Validate.Plain[Double, NotNaN] = fromIsNaN(_.isNaN)
+
+    def fromIsNaN[A](isNaN: A => Boolean): Validate.Plain[A, NotNaN] =
+      Validate.fromPredicate(x => !isNaN(x), x => s"$x != NaN", NotNaN())
```
fthomas: There should be parenthesis in the string representation of this predicate.
```suggestion
      Validate.fromPredicate(x => !isNaN(x), x => s"($x != NaN)", NotNaN())
```
Other predicates do this, too.

dcastro:
Thanks for the feedback @fthomas! I've pushed another commit addressing the comments :)

PR #610, opened 2019-02-01
 Move scala-xml related code to separate module

howyp:
I'm not sure what the naming of this module should be - `scala-xml`, `scalaXml` or `scalaxml`? I tried including the hyphenated version but that seemed to cause problems in the build. Feel free to suggest an alternative to the naming and/or the package layout in the new module!

Note that the `core` module still depends on `scala-xml` because the compiler does - see https://github.com/scala/scala-xml/wiki/FAQ


build.sbt
```diff
@@ -262,6 +262,16 @@ lazy val shapelessJVM = shapeless.jvm
 lazy val shapelessJS = shapeless.js
 lazy val shapelessNative = shapeless.native
 
+lazy val scalaxml = myCrossProject("scalaxml")
+  .dependsOn(core % "compile->compile;test->test")
+  .settings(
+    libraryDependencies += "org.scala-lang.modules" %% "scala-xml" % scalaXmlVersion
```
fthomas: If scala-xml is available for Scala.js and Scala Native, shouldn't this use `%%%`?

howyp:
Sadly, it looks like there's no scala-xml for native:

```
sbt.librarymanagement.ResolveException: unresolved dependency: org.scala-lang.modules#scala-xml_native0.3_2.11;1.1.1: not found
```

howyp:
Ok, looks like this is GTG. Merge at will!

PR #612, opened 2019-02-03
Add type param to autoUnwrap to improve type inference

### LINKED ISSUE #608: autoUnwrap does not kick in when refined type is used in expression
The `eu.timepit.refined.auto.autoUnwrap` implicit conversion does not work when used in expressions.

Example :
```scala
import eu.timepit.refined.api.Refined
import eu.timepit.refined.numeric.Positive
import eu.timepit.refined.auto._

type PositiveInt = Int Refined Positive

val i1: PositiveInt = 5
val i2: PositiveInt = 5

val i3: Int = i1 + i2
```

The last line complains about the + function not being defined for PositiveInt. The autoUnwrap function  work when used directly : 
```scala
val i3: Int = autoUnwrap(i1) + autoUnwrap(i2)
```

In comparison a simpler definition works fine, for example this one :
```scala
@inline implicit def refinedType2Base[Predicate, Base](refinedType: Base Refined Predicate): Base = refinedType.value
```

It would be great if the autoUnwrap def would be called here, as calling `.value`on each Refined type is cumbersome.

fthomas:
closes #608

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/auto.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/auto.scala
@@ -53,7 +53,7 @@ object auto {
    * Note: This conversion is not needed if `F[T, _] <: T` holds (which
    * is the case for `shapeless.tag.@@`, for example).
    */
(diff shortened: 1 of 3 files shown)


fthomas:
I verified that the added test fails to compile prior to this change.

PR #614, opened 2019-02-05
Make scalaxml module JVM-only

fthomas:
This makes the scalaxml module only available for the JVM because
scala-xml is not available for Scala Native and fails with linking
errors on Scala.js:
```
Referring to non-existent class javax.xml.parsers.SAXParserFactory$
   called from scala.xml.factory.XMLLoader.parser()javax.xml.parsers.SAXParser
   called from scala.xml.XML$.parser()javax.xml.parsers.SAXParser
   called from scala.xml.factory.XMLLoader.loadString(java.lang.String)scala.xml.Node
   called from scala.xml.XML$.loadString(java.lang.String)scala.xml.Node
   called from eu.timepit.refined.scalaxml.string$Xml$.$$anonfun$xmlValidate$1(java.lang.String)scala.xml.Elem
   called from eu.timepit.refined.scalaxml.string$Xml$.xmlValidate()eu.timepit.refined.api.Validate
```
See also #610.
/cc @howyp


howyp:
Hmmm - did I mess something up with my PR? Sorry if I did.

But if I'm reading this PR right, does it mean that the scala-xml related code that was in `core` wouldn't have worked on JS/native?

fthomas:
No, your PR was fine. It preserved what was already in the core module and since the core was available for Scala.js and Scala Native it made sense to also build that module for all three platforms.

But you're right, the scala-xml code never worked on JS and native which was the reason for the tests being JVM-only. The reason why it grew this way, was that the scala-xml code was there before we cross-built for other platforms. With your PR and this PR we're being honest again that the XML stuff is only available on the JVM.

Issue #527, opened 2018-06-11
refined-cats: Semigroup for numeric types


kubukoz:
AFAICS it should be possible to define a `Semigroup[A Refined Positive]` for any `A : Semigroup`, as well as for `Semigroup[A Refined Negative]` (adding two negative numbers still gives a negative number).

WDYT about adding something like that to the refined-cats module?

kubukoz:
There could also be semigroups for `Greater[A]` for the same `A`. Ideally given `A < B` we should be able to call `|+|` on values `X Refined Greater[A]` and `Y Refined Greater[B]` and get a `Z Refined Greater[A]`, etc.

fthomas:
I don't think it is possible to derive a `Semigroup[A Refined Positive]` from any `Semigroup[A]`. Here is a contrived counterexample:
```scala
implicit val sg = new Semigroup[Int] { 
  def combine(a: Int, b: Int): Int = -1
}
```
`sg.combine` always returns a negative `Int` and I don't see way to derive a `Semigroup[Int Refined Positive]` from `sg` that only produces positives `Int`s in a generic way.

But adding some specific semigroups / monoids for addition of integral types (e.g. `PosInt` with `+` or `NonNegLong` with `(+, 0)`) should be fine.

kubukoz:
I see... then let's add some instances for actual types :) I'll come up with some PR when I have time.

dm-tran:
I tried to implement some Semigroup instances and tested them using [cats-laws](https://typelevel.org/cats/typeclasses/lawtesting.html).

```
def getPosSemigroup[A: Semigroup](implicit v: Validate[A, Positive]): Semigroup[A Refined Positive] =
  Semigroup.instance{ (x, y) =>
    val combined: A = Semigroup[A].combine(x.value, y.value)
    refineV[Positive](combined).right.get
  }

implicit val posByteSemigroup: Semigroup[PosByte] = getPosSemigroup[Byte]
implicit val posShortSemigroup: Semigroup[PosShort] = getPosSemigroup[Short]
implicit val posIntSemigroup: Semigroup[PosInt] = getPosSemigroup[Int]
implicit val posLongSemigroup: Semigroup[PosLong] = getPosSemigroup[Long]
implicit val posFloatSemigroup: Semigroup[PosFloat] = getPosSemigroup[Float]
implicit val posDoubleSemigroup: Semigroup[PosDouble] = getPosSemigroup[Double]
```

[...]

PR #618, opened 2019-03-09
add semigroup and monoid instances for numeric types

dm-tran:
This merge request partially fixes https://github.com/fthomas/refined/issues/527. It adds:

- semigroup instances for Positive and Negative numeric types (Byte, Short, Int, Long, Float and Double)
- monoid instances for NonPosFloat, NonPosDouble and NonNegative numeric types (Byte, Short, Int, Long, Float and Double) 
- law tests for these instances. Law tests for Float and Double are commented (as done in [cats tests](https://github.com/typelevel/cats/blob/v1.6.0/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala#L218)), because floating point addition is not associative.

As discussed in https://github.com/fthomas/refined/issues/527, the instances overflow for Integral types.

FYI, I tried adding monoid instances for NonPositive Integral types, but they are not associative:
```
  implicit val nonPosByteMonoid: Monoid[NonPosByte] = getNonPosIntegralMonoid[Byte]
  implicit val nonPosShortMonoid: Monoid[NonPosShort] = getNonPosIntegralMonoid[Short]
  implicit val nonPosIntMonoid: Monoid[NonPosInt] = getNonPosIntegralMonoid[Int]
  implicit val nonPosLongMonoid: Monoid[NonPosLong] = getNonPosIntegralMonoid[Long]

  private def getNonPosIntegralMonoid[A: Integral: Monoid: NegShift](
      implicit v: Validate[A, NonPositive]
  ): Monoid[A Refined NonPositive] = new Monoid[A Refined NonPositive] {
    override def empty: A Refined NonPositive = refineV[NonPositive](Monoid[A].empty).right.get

    override def combine(
        x: A Refined NonPositive,
        y: A Refined NonPositive
    ): A Refined NonPositive = {
      val combined: A = x.value |+| y.value

      refineV[NonPositive](combined).getOrElse {
        val result: A = NegShift[A].shift(combined) |+| implicitly[Integral[A]].one
        refineV[NonPositive](result).right.get
      }
    }
  }
```

Example of failed test for Monoid[NonPosByte] : 
```
[info] - Monoid[NonPosByte].monoid.associative *** FAILED ***
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
[info]    (Discipline.scala:14)
[info]     Falsified after 25 successful property evaluations.
[info]     Location: (Discipline.scala:14)
[info]     Occurred when passed generated values (
[info]       arg0 = -1,
[info]       arg1 = -128,
[info]       arg2 = -128
[info]     )
[info]     Label of failing property:
[info]       Expected: -1
[info]   Received: -128
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,9 +100,12 @@ lazy val benchmark = project
   .settings(noPublishSettings)
 
 lazy val cats = myCrossProject("cats")
-  .dependsOn(core % "compile->compile;test->test")
+  .dependsOn(core % "compile->compile;test->test", scalacheck_1_13 % Test)
   .settings(
-    libraryDependencies += "org.typelevel" %%% "cats-core" % catsVersion,
+    libraryDependencies ++= Seq(
+      "org.typelevel" %%% "cats-core" % catsVersion,
+      "org.typelevel" %% "cats-testkit" % catsVersion % Test
+    ),
     initialCommands += s"""
       import $rootPkg.cats._
     """
--- a/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/Shift.scala
+++ b/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/Shift.scala
@@ -1,0 +1,43 @@
+package eu.timepit.refined.cats
+
+/**
+ * Typeclass to shift Negative values to Non Negative values.
+ */
+private[cats] trait NonNegShift[T] extends Serializable {
+  def shift(t: T): T
+}
+
+private[cats] object NonNegShift {
+  def apply[T](implicit ev: NonNegShift[T]): NonNegShift[T] = ev
+
+  def instance[T](function: T => T): NonNegShift[T] = new NonNegShift[T] {
+    def shift(t: T): T = function(t)
+  }
+
+  // Instances
+  implicit val byteNonNegShift: NonNegShift[Byte] = instance(t => (t & Byte.MaxValue).toByte)
+  implicit val shortNonNegShift: NonNegShift[Short] = instance(t => (t & Short.MaxValue).toShort)
+  implicit val intNonNegShift: NonNegShift[Int] = instance(t => t & Int.MaxValue)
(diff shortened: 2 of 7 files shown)


dm-tran:
The build fails with scala 2.13.0-M5 because of the scalacheck dependency.

I tried to replace 
```
 lazy val cats = myCrossProject("cats")
  .dependsOn(core % "compile->compile;test->test", scalacheck_1_13 % Test)
```
by
```
 lazy val cats = myCrossProject("cats")
  .dependsOn(
    core % "compile->compile;test->test",
    CrossVersion.partialVersion(scalaVersion.value) match {
      case Some((2, minor)) if minor <= 12 => scalacheck_1_13 % Test
      case _ => scalacheck % Test
    }
  )
```

But the following error is raised when launching sbt : 
```
error: `value` can only be used within a task or setting macro, such as :=, +=, ++=, Def.task, or Def.setting.
```

@fthomas I'm not familiar with cross building. Do you have any idea of how I could fix the build for scala 2.13.0-M5 ?

fthomas:
Oh no... I've no immediate idea how to fix this. :-(

bplommer:
Looks like the Scala 2.13 version bump fixed this build 🙂

dm-tran:
> Looks like the Scala 2.13 version bump fixed this build 🙂

Actually, module `refined-cats` isn't built for scala 2.13 for the time being, because of commit https://github.com/fthomas/refined/commit/155e7b7a2f97eb99f12f2501b5a889c690f9c481

Cats will add support for scala 2.13 in version 2.0.0, which isn't yet released. Supporting scala 2.13 for refined-cats means migrating from cats 1.6.1 to cats 2.0.0.

This MR could be merged, so that these new semigroup and monoid instances are available for refined-cats in scala 2.11 and 2.12. @fthomas what do you think ?

According to https://github.com/typelevel/cats/blob/master/ROADMAP_2019.md#q1, Cats 2.0.0 will migrate from scalacheck 1.13 to scalacheck 1.14. So, migrating from cats 1.6.1 to cats 2.0.0 should simplify the build.

PR #638, opened 2019-05-30
Add FOLIO to list of adopters

rider-yi:
We are using refined in our financial systems :)


fthomas:
Nice! Thanks @ryder-yi!

PR #642, opened 2019-06-04
Cats validatenec

BalmungSan:
I think it may be useful _(at least it is for me 😅)_.

Uhm, I just realized that the rename from `validate` to `validateNel` would break existing code...
Would you want me to leave it as `validate`, or keep both `validate` & `validateNel` _(maybe in a future deprecate the first?)_.


fthomas:
:+1: 

I'm fine with adding `validateNel` alongside `validate`. When we bump the version to 0.10, we can then mark `validate` as deprecated and then remove it in 0.11 or later.

BalmungSan:
Done, I hope this time everything should pass.

fthomas:
Merging, thanks @BalmungSan!

PR #713, opened 2020-01-09
Merge series/0.10 into master

fthomas:


fthomas:
Before merging, this should happen on the `series/0.10` branch before:
* [ ] Remove deprecated stuff in `RefType` and `generic`
* [ ] Move support for `shapeless.tag.@@` to shapeless module
* [ ] Use `ValueOf` instead of `Witness` and proper literals for 2.13
* [ ] Use `Int` literals for `Positive` and other predicates (https://github.com/fthomas/refined/pull/841#issuecomment-699502818)
* [ ] Deprecate support for `shapeless.Nat` (https://github.com/fthomas/refined/issues/763)

PR #724, opened 2020-01-31
WIP: Validator trait modification proposal

strokyl:
While doing this PR for Tapir: https://github.com/softwaremill/tapir/pull/405.
I realize it would be great to be able to serialize a validator without having to run it again a concrete value.

For that I think we could just use showRepr by replacing `t` by "x".

In order to do that, it would need a lot of change.
Before doing all of them, here a work in progress PR.

Tell me what you think about it.


strokyl:
Up!

PR #756, opened 2020-04-08
Fix #755 - expose Implies[P,C] variation of Inference[P,C]

### LINKED ISSUE #755: RFE: A version of Inference[P, C] that fails for implicit chaining
I could use a version of `Inference` that always fails to manifest. For example, the following inference "fails" but it doesn't fail to manifest, it just has a `false` value:
```scala
scala> implicitly[Inference[Greater[-1], Greater[0]]]
res7: eu.timepit.refined.api.Inference[eu.timepit.refined.numeric.Greater[-1],eu.timepit.refined.numeric.Greater[0]] = Inference(false,greaterInference(-1, 0))
```
 This works for the casting in `InferMacro` and `autoInfer` but it's limited in how it can be used in chained implicits.

I could use an alternative that will fail like so:
```scala
scala> implicitly[ChainableInference[Greater[-1], Greater[0]]] 
   error: could not find implicit value ...
```
Given the current design, I think that another similar macro which returns a chainable inference type but only if `inference.notValid` is false, instead of just doing an implicit value conversion.

erikerlandson:
Fixes #755. `Implies[P, C]` has essentially same semantic as `Inference[P, C]` except that instead of manifesting with a `false` conclusion, it will NOT manifest, and so it can be used effectively in chained implicit definitions.

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/api/Inference.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/api/Inference.scala
@@ -33,3 +33,27 @@ object Inference {
   ): Inference[P, C] =
     Inference(i1.isValid && i2.isValid, show.format(i1.show, i2.show))
 }
+
+/**
+ * Similar to `Inference[P, C]` but will not implicitly manifest if `C` cannot be
+ * inferred from `P`.
+ *
+ * It is intended to be used with chained implicit definitions that require proof that `P ==> C`
+ */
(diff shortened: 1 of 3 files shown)


erikerlandson:
Odd, the travis dash shows all tests as having passed but the GH summary is frozen

erikerlandson:
I'm going to close this, provisionally, and take a stab at a pull-req for macro-less `==>`

PR #758, opened 2020-04-14
Draft macroless ==>

erikerlandson:
A candidate fix for #755 that refactors `==>` to be macroless: if a valid inference cannot be proven, then `P1 ==> P2` will fail to manifest. It no longer has a boolean field that is checked via macro.

Introduces a compile-time version of `Refined` called `RefinedLT`, which tests a literal type `L`'s value against a corresponding predicate `P`. So the use of macros has been pushed from `==>` to `RefinedLT`, where the macros are used to extract literal-type value and run the validator against it.

This PR is currently a bit rough - a few unit tests are failing, or failing to compile, which appear to work OK when I run them in the REPL.  I also hacked around the old `.isValid` based unit-testing using a shim that made kittens cry.


erikerlandson:
I have this passing validateJVM locally, but it has to build against a version of singleton-ops with https://github.com/fthomas/singleton-ops/issues/134, I'm using a publishLocal against head of master.

fthomas:
I applause the macroless `==>`!

One quick question: Is `RefinedLT` only required to implement `equalValidateInference` or are there other reasons or instances it is required for?

erikerlandson:
@fthomas that's right, I only implemented `RefinedLT` to solve `equalValidateInference`. It seemed possibly useful as an implicit tool in its own right. For example, expressing refined-like constraints on type variables. But that's purely speculative.

erikerlandson:
It seems like it could be possible for `refined` to go completely macro-less, if (for example) `refineMV[Positive](5)` was instead written as `refineMV[Positive, 5]`.

This would involve the compile-time `refineMV` leaning totally on singleton-ops, and expressions of the (pseudo)form `implicit valid: OpAuxBoolean[P[V], true]`.  In that respect, all the macros get pushed to singleton-ops.  https://github.com/fthomas/singleton-ops/issues/134 would probably get singleton-ops close to supporting all of refined, maybe not all the way.  It could put `refined` in a better position for scala 3, and localize all the macro porting to singleton-ops.

In pre-2.13 scala this would not be as nice:  `refineMV[Positive, W.`5`.T]`, but the future is first-class literal types.

erikerlandson:
Probably obvious from the PR diff, and my comments, but worth calling out that this does add `singleton-ops` as a new dependency.

fthomas:
My thoughts with regards to `equalValidateInference`: If we accept that the `Inference#isValid` field was a mistake, then `equalValidateInference` was a mistake too. The instance was added without explanation in https://github.com/fthomas/refined/commit/f1812d88042b143b826b8ce423d13501a6c1a197 and is in the library since version 0.1.0. I guess my reasoning for adding it was something like this: if `Inference` requires a `Boolean` we can go from any value `T Refined Equal[U]` (which has the value `U`) to `T Refined P` by applying `P` to `U` to obtain that `Boolean`. So that instance either requires the questionable `isValid` field or another macro (`RefinedLT`). I think if we want to put refined in a better position for Scala 3 we should not trade one macro for another and remove `equalValidateInference` altogether.

WDYT, @erikerlandson?

erikerlandson:
>  I think if we want to put refined in a better position for Scala 3 we should not trade one macro for another and remove equalValidateInference altogether.

@fthomas  I have no use cases that would need the `equalValidateInference` rule. If you are OK with removing it, it seems like a good trade-off for truly reducing the number of macros.

erikerlandson:
I'm interested in the possibility of somehow unifying run-time` (`refineV`) with compile-time `refineMV`, via the `Validate` implicit system.  I am not sure if it's feasible.  Maybe it could be done by adding some kind of type `Pred[L]` to `Validate`, where one could compile-time test `OpAuxBoolean[Pred[L], true]` in the case of `refineMV`, and so compile-time and run-time tests are built-up in parallel via `Validate`. 

I have also been experimenting with the idea of somehow binding literal type from a function that requires some kind of constant-expression as parameter. In fake scala, something in the vein of:
```scala
def refineMV[P, T <: Singleton](constExpr: T)(implicit vt: Validate[P, T] ...)
```
Not sure if scala allows that to work.  If not, a form like `refineMV[P, V]` would be a possible alternative to the current `refineMV[P](v)`

erikerlandson:
just as a use-case reference, here is how I'm applying `==>` to constrain operations in type sound ways:
https://github.com/erikerlandson/coulomb/blob/refined/coulomb-refined/src/main/scala/coulomb/refined/package.scala#L154

erikerlandson:
@fthomas I removed RefinedLT and equalValidateInference.  There were two impacts.

1) this inference is no longer possible:
```scala
  property("Exists[A] ==> Exists[B]") = secure {
    Inference[Contains[W.`'5'`.T], Exists[Digit]].isValid
  }
```

2) this assignment no longer compiles
```scala
    val a: Char Refined Equal[W.`'0'`.T] = '0'
    val b: Char Refined Digit = a
```

PR #760, opened 2020-04-15
Add new NonEmptyFiniteString type

Hjdskes:
This type is a conjunction of the existing FiniteString and NonEmptyString types.

In our codebase at work, we often have to deal with strings that are both non-empty and finite of length. I thought this could be a useful addition to refined, rather than just having it privately in our codebase.

I'll comment down below on things that I am not sure of or that are currently broken that I don't know how to fix.

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
@@ -118,12 +118,14 @@ trait StringTypes {
   final type NonEmptyString = string.NonEmptyString
   final val NonEmptyString = string.NonEmptyString
 
-  final type NonEmptyFiniteString[N] = string.NonEmptyFiniteString[N]
-  final val NonEmptyFiniteString = string.NonEmptyFiniteString
-
   final type TrimmedString = string.TrimmedString
   final val TrimmedString = string.TrimmedString
 
   final type HexString = string.HexString
   final val HexString = string.HexString
+}
+
+trait StringTypesBinCompat1 {
+  final type NonEmptyFiniteString[N] = string.NonEmptyFiniteString[N]
+  final val NonEmptyFiniteString = string.NonEmptyFiniteString
 }
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
@@ -1,8 +1,10 @@
 package eu.timepit.refined.types
 
 import eu.timepit.refined.api.{Refined, RefinedType, RefinedTypeOps}
-import eu.timepit.refined.collection.{MaxSize, NonEmpty}
+import eu.timepit.refined.collection.{MaxSize, NonEmpty, Size}
+import eu.timepit.refined.numeric.Interval
 import eu.timepit.refined.string.{HexStringSpec, Trimmed}
+import shapeless.Nat._1
 import shapeless.Witness
 
 /** Module for `String` refined types. */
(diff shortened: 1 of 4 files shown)


modules/core/shared/src/main/scala/eu/timepit/refined/string.scala
```diff
@@ -244,6 +247,17 @@ object string extends StringInference {
       )
   }
 
+  object NonEmptyFinite {
+    implicit def nonEmptyFiniteValidate[N <: Int](
```
Hjdskes, line 251: I'm not sure if it is possible to derive this from pre-existing `Validate` instances for `NonEmpty` and `MaxSize[N]`. I tried without it but that didn't seem to work, unless I was doing something wrong.

modules/scalacheck/shared/src/test/scala/eu/timepit/refined/scalacheck/StringArbitrarySpec.scala
```diff
@@ -26,4 +31,6 @@ class StringArbitrarySpec extends Properties("StringArbitrary") {
   property("FiniteString[10]") = checkArbitraryRefinedType[FiniteString[W.`10`.T]]
 
   property("Size[Equal[8]]") = checkArbitraryRefinedType[String Refined Size[Equal[W.`8`.T]]]
+
+  property("NonEmptyFiniteString[10]") = checkArbitraryRefinedType[NonEmptyFiniteString[W.`10`.T]]
```
Hjdskes, line 35: This fails to compile due to diverging implicit resolution:

```scala
/Users/hjdskes/github/refined/modules/scalacheck/shared/src/test/scala/eu/timepit/refined/scalacheck/StringArbitrarySpec.scala:35:67: diverging implicit expansion for type org.scalacheck.Arbitrary[(eu.timepit.refined.collection.NonEmpty, eu.timepit.refined.collection.MaxSize[Int(10)])]
[error] starting with method arbTuple2 in trait ArbitraryArities
[error]   property("NonEmptyFiniteString[10]") = checkArbitraryRefinedType[NonEmptyFiniteString[W.`10`.T]]
```

Do I need to define a new method in `scalacheck/string.scala` to get this to work?

fthomas: Maybe it compiles if `NonEmptyFiniteString` is defined as `String Refined Size[Interval.Closed[_1, N]]`.

modules/core/shared/src/main/scala/eu/timepit/refined/string.scala
Hjdskes: Hmm yea.. seems like it was a leftover requirement from earlier experiments maybe. In any case, it's gone now :)

modules/scalacheck/shared/src/test/scala/eu/timepit/refined/scalacheck/StringArbitrarySpec.scala
Hjdskes: Yep, it does!

modules/core/shared/src/main/scala/eu/timepit/refined/types/string.scala
```diff
@@ -61,6 +62,39 @@ object string {
 
   object NonEmptyString extends RefinedTypeOps[NonEmptyString, String]
 
+  /** A `String` that is not empty with length less than or equal to `N`. */
+  type NonEmptyFiniteString[N] = String Refined And[NonEmpty, MaxSize[N]]
```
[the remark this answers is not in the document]
Hjdskes: Thanks, that's the link I was missing. My initial code on Gitter was wrong because it had an upper bound on the type variable `N`, I suppose.

Hjdskes:
I've pushed a new version defined as `type NonEmptyFiniteString[N] = String Refined Size[Interval.Closed[_1, N]]`. It seems to solve all issues, fingers crossed for the build :)

Hjdskes:
Argh, there was a leftover predicate case class so I pushed a new version.

Would it make sense to add some kind of `truncate` method to `NonEmptyFiniteStringOps`, similar to `FiniteString`'s? The only downside is that this would have to perform an emptiness check, unless we make it `NonEmptyString => NonEmptyFiniteString` rather than `String => Either[String, NonEmptyFiniteString]`.

Hjdskes:
The Scala 2.12.11 build seems to fail on MiMa. How do you prefer me to fix that?

fthomas:
The MiMa errors should go away if these
```scala
  final type NonEmptyFiniteString[N] = string.NonEmptyFiniteString[N]
  final val NonEmptyFiniteString = string.NonEmptyFiniteString
```
are not added to the `trait StringTypes` but to a new `trait StringTypesBinCompat1` similar to what was done in https://github.com/fthomas/refined/commit/5ac1c1b35f1466f35b3ee838cc03727c2c3186bb. `StringTypesBinCompat1` then needs to extend a new `trait AllTypesBinCompat3` which is then mixed into the [`types.all` object](https://github.com/fthomas/refined/blob/5d385acbf0ef9f3c807559019437e0db1dd80293/modules/core/shared/src/main/scala/eu/timepit/refined/types/all.scala).

fthomas:
Do you need a `truncate` method for `NonEmptyFiniteString`? If not I wouldn't add it now.

I think it was added to `FiniteString` because there is only one obvious way to implement it and it allows to convert _any_ `String` to a `FiniteString`

Hjdskes:
Yea, makes sense. If we have a good use-case for it I'll add it later with a new PR.

I've pushed a commit that fixes the binary incompatibility. Let me know if you want me to squash that, or feel free to do that yourself during the merge.

fthomas:
You don't need to squash the commits, I'll merge both.

Thanks for your contribution!

Hjdskes:
Thank you for the quick review and assistance! :)

PR #761, opened 2020-04-15
Adding neotypes to the external modules list

BalmungSan:
[**neotypes**](https://neotypes.github.io/neotypes/) is a Scala lightweight, type-safe, asynchronous driver for neo4j. And we provide a module to allow inserting and reading refined types.

And we would like to be added on the external modules list.


fthomas:
Thanks @BalmungSan!

PR #772, opened 2020-05-08
Add Inference instance for Size[Interval.Closed[1, N]] ==> NonEmpty

Hjdskes:
This allows, amongst others, the inference of `NonEmptyString` from `NonEmptyFiniteString`.

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/collection.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/collection.scala
@@ -8,7 +8,7 @@ import eu.timepit.refined.generic.Equal
 import eu.timepit.refined.internal.Resources
 import eu.timepit.refined.numeric.{GreaterEqual, Interval}
 import shapeless.Witness
-import shapeless.nat._0
+import shapeless.nat.{_0, _1}
 
 /** Module for collection predicates. */
 object collection extends CollectionInference {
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/eu/timepit/refined/collection.scala
```diff
@@ -342,4 +342,7 @@ private[refined] trait CollectionInference {
 
   implicit def sizeInference[A, B](implicit p1: A ==> B): Size[A] ==> Size[B] =
     p1.adapt("sizeInference(%s)")
+
+  implicit def sizeNonEmpty[N]: Size[Interval.Closed[_1, N]] ==> NonEmpty =
```
fthomas, line 346: For consistency with the other instances:
```suggestion
  implicit def sizeNonEmptyInference[N]: Size[Interval.Closed[_1, N]] ==> NonEmpty =
```

Hjdskes:
I've applied your suggestion from GitHub. Feel free to squash it :)

Hjdskes:
Awesome!

fthomas:
I was also happy that this instance works! 

FTR: It implies `NonEmpty` for every predicate `Size[A]` where the predicate `A` implies `>= 1`.

Hjdskes:
Does this PR need anything else? :)

fthomas:
No, I think it is ready. I just wanted to note that the conversion from `NonEmptyFiniteString` to `NonEmptyString` involves an `Inference` with three nested instances as can been seen here:
```scala
scala> val i = Inference[Size[Interval.Closed[_1, W.`5`.T]], NonEmpty]
i: Inference[Size[Interval.Closed[_1,Int(5)]],NonEmpty] = Inference(true,sizeGreaterEqual1NonEmptyInference(conjunctionEliminationL(minimalTautology)))
```

PR #787, opened 2020-07-10
upgrading pureconfig to 0.13.0

BusyByte:
resolves: https://github.com/fthomas/refined/pull/783

We are blocked on upgrading pureconfig to 0.13.0 in our application until refined is released with an upgraded dependency.


BusyByte:
@fthomas let me know if this looks ok or if there is anything else which needs changed.

fthomas:
refined 0.9.15 is out now and contains this update.

Issue #702, opened 2019-12-03
PrivateNetwork type does not match the expected IP addresses


steinybot:
Was there a particular reason behind the RFCs that were chosen in https://github.com/fthomas/refined/pull/356?

I found that `PrivateNetwork` did not match what I expected it to.

The addresses in [rfc5737](https://tools.ietf.org/html/rfc5737#section-4) seem like they should never be routed to:
> 4.  Operational Implications
>
>    Addresses within the TEST-NET-1, TEST-NET-2, and TEST-NET-3 blocks
>    SHOULD NOT appear on the public Internet and are used without any
>    coordination with IANA or an Internet registry [RFC2050].  Network
>    operators SHOULD add these address blocks to the list of non-
>    routeable address spaces, and if packet filters are deployed, then
>    this address block SHOULD be added to packet filters.
>
>    These blocks are not for local use, and the filters may be used in
>    both local and public contexts.

There is also a bug in `Rfc1918ClassBPrivateSpec`. It should start at `172.16.0.0` not `172.15.0.0`. https://tools.ietf.org/html/rfc1918#section-3

I'm by no means an expert but it seems like adding types for [rfc6890](https://tools.ietf.org/html/rfc6890) would be a better way to go.

Without digging into each RFC I suspect what most people would want is any address which is listed in rfc6890 as a valid destination IP address and maybe the loopback addresses.
>Destination - A boolean value indicating whether an address from
      the allocated special-purpose address block is valid when used as
      the destination address of an IP datagram that transits two
      devices.

Thoughts?

jbenoit2011:
IP adresses in Rfc5737 are **reserved** for documentation usage. Rfc6890 list address blocks for different purposes, except public routing, like benchmarking, documentation, private routing, ...

IMHO, address blocks not involved in private or public networks, if they come to be managed by this library, should be isolated in a separated object, something like `SpecificPurposesIpAddressBlocks`.

But the only use case I see is to exclude those specific-purpose IPs, you can already do that by combining both public and private types. So I don't think it worth the effort just for the completeness.

PR #796, opened 2020-08-13
Fix Rfc1918ClassBPrivateSpec incorrect matches

### LINKED ISSUE #795: Rfc1918ClassBPrivateSpec incorrect matches
[Rfc1918ClassBPrivateSpec](https://github.com/fthomas/refined/blob/e64f11c6f4bcb322ceac76da4011c7e993838f5b/modules/core/shared/src/main/scala/eu/timepit/refined/types/net.scala#L79) matches IPs out of RFC range.
For example, `172.222.171.197` will match.
This is caused by a missing escaping backslash before the latest dot in the regex.
Actually, the regex is `^172\.(15|16|17|18|19|20|21|22|23|24|25|26|27|28|29|30|31).+`, the latest dot is considered as a wildcard character because not escaped, so any IP with a second digit starting with characters 15 to 31 will match.

Adding a backslash before, solve the issue: `^172\.(15|16|17|18|19|20|21|22|23|24|25|26|27|28|29|30|31)\.+`

Here is a scastie to test: https://scastie.scala-lang.org/ktfISeSZQNG9iLt7MvUg0g

May I send a PR?

Thanks.

jbenoit2011:
Fix #795

### DIFF
--- a/modules/core/shared/src/main/scala/eu/timepit/refined/types/net.scala
+++ b/modules/core/shared/src/main/scala/eu/timepit/refined/types/net.scala
@@ -76,7 +76,9 @@ object net {
       IPv4 And StartsWith[W.`"10."`.T]
 
     type Rfc1918ClassBPrivateSpec =
-      IPv4 And MatchesRegex[W.`"^172\\\\.(15|16|17|18|19|20|21|22|23|24|25|26|27|28|29|30|31).+"`.T]
+      IPv4 And MatchesRegex[
+        W.`"^172\\\\.(15|16|17|18|19|20|21|22|23|24|25|26|27|28|29|30|31)\\\\..+"`.T
+      ]
 
     type Rfc1918ClassCPrivateSpec =
       IPv4 And StartsWith[W.`"192.168."`.T]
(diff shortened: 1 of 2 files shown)


jbenoit2011:
A test failed, but it does not seem related to my changes:
```
[info] + RefTypeRead.load success: OK, proved property.
1179Error: Option --foo failed when given '0'. Predicate failed: (0 > 0).
1180Usage: tests [options]
1181
1182  -f, --foo <value>  foo is a positive integer property
1183[info] + RefTypeRead.load failure (predicate): OK, proved property.
1184Error: Option --foo expects a number but was given 'abc'
1185Usage: tests [options]
1186
1187  -f, --foo <value>  foo is a positive integer property
```
It already happened on master.

fthomas:
I think the error that failed the build is:
```
[warn] /home/travis/build/fthomas/refined/modules/core/shared/src/main/scala/eu/timepit/refined/types/net.scala isn't formatted properly!
[warn] /home/travis/build/fthomas/refined/modules/core/shared/src/main/scala/eu/timepit/refined/types/net.scala isn't formatted properly!
...
[error] 1 files must be formatted
[error] 1 files must be formatted
[error] (coreJS / Compile / scalafmtCheck) 1 files must be formatted
[error] (coreJVM / Compile / scalafmtCheck) 1 files must be formatted
```
Running `sbt scalafmtAll` should fix the formatting.

jbenoit2011:
Didn't know about `scalafmtAll` command, thanks for telling me.
Everything seems ok now.
When do you plan the next release including this fix?
Thanks.

fthomas:
I can make a release with this fix soon. But before I do, it may be worth looking into another issue with `Rfc1918ClassBPrivateSpec ` that is mentioned in #702:

> There is also a bug in Rfc1918ClassBPrivateSpec. It should start at 172.16.0.0 not 172.15.0.0. https://tools.ietf.org/html/rfc1918#section-3

WDYT @jbenoit2011?

jbenoit2011:
I just pushed a patch to fix the start range of rfc1918 class b ips issue described in #702

PR #832, opened 2020-09-22
Use ValueOf instead of Witness.Aux on Dotty

fthomas:


fthomas:
Closing in favor of #833.

PR #854, opened 2020-10-12
Add take method to NonEmptyList and NonEmptyVector

zaneli:
I want to add `take` method to `NonEmptyList` and `NonEmptyVector` for convenience.

### DIFF
--- a/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/syntax.scala
+++ b/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/syntax.scala
@@ -25,7 +25,7 @@ trait CatsNonEmptyListSyntax {
     def refinedSize: PosInt = PosInt.unsafeFrom(nel.size)
 
     def take(n: PosInt): NonEmptyList[A] =
-      NonEmptyList.fromListUnsafe(nel.toList.take(n.value))
+      NonEmptyList(nel.head, nel.tail.take(n.value - 1))
   }
 }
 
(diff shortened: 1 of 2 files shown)


modules/cats/shared/src/main/scala/eu/timepit/refined/cats/syntax.scala
```diff
@@ -23,5 +23,15 @@ trait CatsRefinedTypeOpsSyntax {
 trait CatsNonEmptyListSyntax {
   implicit class CatsNonEmptyListRefinedOps[A](nel: NonEmptyList[A]) {
     def refinedSize: PosInt = PosInt.unsafeFrom(nel.size)
+
+    def take(n: PosInt): NonEmptyList[A] =
+      NonEmptyList.fromListUnsafe(nel.toList.take(n.value))
```
fthomas, line 28: This could be made a bit more performant by not calling `toList`:
```suggestion
      NonEmptyList(nel.head, nel.tail.take(n.value - 1))
```

fthomas (approved):
LGTM. Many thanks, @zaneli! I added one comment about the performance of `take` for `NonEmptyList`.

modules/cats/shared/src/main/scala/eu/timepit/refined/cats/syntax.scala
zaneli: Thank you for the comment!
I committed this suggestion.

PR #867, opened 2020-11-12
Check that Arbitrary[NonEmptyString] does not discard values

fthomas:


fthomas:
The build fails with:
```
[info] ! ZeroDiscardRatioSpec.NonEmptyString: Gave up after only 0 passed tests. 1 tests were discarded.
```
This is due to https://github.com/typelevel/scalacheck/issues/708 which will be fixed in the next ScalaCheck version. I've verified locally that this test passes with ScalaCheck 1.14.3.

PR #881, opened 2020-12-04
Depend on ScalaCheck Scala 3 jar

guymers:
Depend on the ScalaCheck Scala 3 jar to avoid requiring downstream
consumers to also use the 2.13 ScalaCheck jar.


fthomas (approved):
LGTM, many thanks @guymers!

PR #887, opened 2020-12-14
Add scallop module

Rogach:
I'm submitting this PR on behalf of [scallop#194](https://github.com/scallop/scallop/issues/194). Scallop is a CLI parser library, similar to scopt. Implementation of refined is also very similar to scopt's implementation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
         uses: codecov/codecov-action@v1
 
       - name: Compress target directories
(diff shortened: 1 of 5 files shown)


Rogach:
Sorry for making such a mess with multiple commits - took me several tries to figure out the proper validation setup. Should I squash&rebase or you will squash when merging?

P.S. Very impressibe build setup, very well thought out :)

fthomas:
Sorry for the late reply @Rogach. The code in this PR looks good but I currently have not enough free time to maintain another integration module. I'm afraid this module needs to live somewhere else.

PR #904, opened 2021-02-02
Adds Validate[F[T], Size[P]] for every F[_]: Foldable.

froth:
This allows refining NonEmptyList et.al. which do not implement Iterable.

For some Foldables there of course exist higher performance implementations. However this implementation is very general and for NonEmptyList / NonEmptyChain at least is very similar with regard of performance.

Concerning the tests: I am not very happy with them because they also test logic outside of my implementation and rely on the String representations of predicates. Ideas welcome.

### DIFF
--- a/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/derivation.scala
+++ b/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/derivation.scala
@@ -32,4 +32,5 @@ trait DerivationInstances {
       gt: G[T]
   ): G[F[T, P]] =
     m.flatMap(gt)(t => m.fromEither(rt.refine[P](t)))
+
 }
--- a/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/unorderedFoldable.scala
+++ b/modules/cats/shared/src/main/scala/eu/timepit/refined/cats/unorderedFoldable.scala
@@ -1,0 +1,39 @@
+package eu.timepit.refined.cats
+
+import cats.UnorderedFoldable
+import cats.syntax.unorderedFoldable._
+import eu.timepit.refined.api.Validate
+import eu.timepit.refined.collection.Size
+import eu.timepit.refined.internal.Resources
+
+/** Module for unorderedFoldable instances. */
+object unorderedFoldable extends UnorderedFoldableInstances
+
+trait UnorderedFoldableInstances {
+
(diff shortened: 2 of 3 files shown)


modules/cats/shared/src/main/scala/eu/timepit/refined/cats/derivation.scala
```diff
@@ -32,4 +35,33 @@ trait DerivationInstances {
...
+        v.showExpr(size(t))
+
+      override def showResult(t: F[T], r: Res): String = {
+        val s = size(t)
+        val nested = v.showResult(s, r.detail.p)
+        Resources.predicateTakingResultDetail(s"size($t) = $s", r, nested)
+      }
+
+      private[this] def size(t: F[T]): Int =
+        t.foldMap(_ => 1)
```
markus1189, line 65: how about `Foldable[F].size(t)`?

froth: yeah, way better, thanks :)

fthomas:
This looks great, thanks @froth! I've two comments:
 1. `Foldable` inherits `size` from `UnorderedFoldable`, so I guess that instance could work for all `UnorderedFoldable`.
 1. I would put the instance in a new `foldable` (or `unorderedFoldable`) object instead. I think that better fits we how organize the predicates and `Validate` instances in the library.

fthomas (approved):
Thanks!

froth:
@fthomas one last question: 

Would it make sense to make the cats package object extend the UnorderedFoldableInstances trait?

This would be the way I would have expeced it to be. "Oh I can't refine that cats.data._ type, let's import refined.cats._"

On the other hand there exist some Foldable instances that are a "little" weird. Like Foldable[Seq] used with an infinite LazyList.

What do you think?

fthomas:
In principle I'm fine with that but I would check if the `Validate[T, Size[P]]` instance in `import eu.timepit.refined.collection._` clashes with the instance in `unorderedFoldable` if one tries to refine a `List` with `Size`. IMO it would be unfortunate if you can't import `refined.collection._` and `refined.cats._` because of that.

froth:
@fthomas  Ah, you are right. It does not clash in "does not compile" but it uses the new Foldable Instance for List if both are imported, which is not what you would expect.

PR #905, opened 2021-02-02
Fix the links to the relevant Scala files

kyusu:
While browsing the README.md I noticed that the links to the Scala files in the "Provided Predicates" section are broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,7 +326,7 @@ page in the wiki.
 
 The library comes with these predefined predicates:
 
-[`boolean`](https://github.com/fthomas/refined/blob/master/modules/core/shared/src/main/scala/eu/timepit/refined/predicates/boolean.scala)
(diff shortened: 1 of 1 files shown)


fthomas:
Thanks for the PR, @kyusu. I broke the links while adding support for Scala 3 where I moved the source files to Scala version specific directories, e.g: https://github.com/fthomas/refined/blob/master/modules/core/shared/src/main/scala-3.0+/eu/timepit/refined/boolean.scala
I'd prefer linking to these files instead of the files in the `predicate` package since the latter contain only forwarders to the real code in the former files.

kyusu:
@fthomas
Thanks for looking at the pull request. I'll change the links!

fthomas (approved):
Thanks @kyusu!

kyusu:
I've changed the links to `boolean`, `numeric` & `generic`.

PR #925, opened 2021-03-10
Use GitHub URLs instead of local paths in SJSIR files

fthomas:
See https://github.com/raquo/Laminar/pull/91 why we're doing this.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -369,7 +369,18 @@ def moduleJsSettings(name: String): Seq[Def.Setting[_]] =
     crossScalaVersions := moduleCrossScalaVersionsMatrix(name, JSPlatform),
     doctestGenTests := Seq.empty,
     mimaFailOnNoPrevious := false,
-    coverageEnabled := false
+    coverageEnabled := false,
+    scalacOptions ++= {
+      if (isDotty.value) Seq()
+      else {
+        val tagOrHash =
+          if (!isSnapshot.value) s"v${version.value}"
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -369,7 +369,18 @@ def moduleJsSettings(name: String): Seq[Def.Setting[_]] =
     crossScalaVersions := moduleCrossScalaVersionsMatrix(name, JSPlatform),
     doctestGenTests := Seq.empty,
     mimaFailOnNoPrevious := false,
-    coverageEnabled := false
+    coverageEnabled := false,
+    scalacOptions ++= {
+      if (isDotty.value) Seq()
```
xuwei-k, line 374: - https://github.com/lampepfl/dotty/blob/95fa9cb9f3a6989ae41bbe047fc977cc33716b87/compiler/src/dotty/tools/dotc/config/ScalaSettings.scala#L130
- https://github.com/lampepfl/dotty/blob/95fa9cb9f3a6989ae41bbe047fc977cc33716b87/compiler/src/dotty/tools/backend/sjs/JSPositions.scala#L20-L30
- https://github.com/scala/scala-collection-compat/blob/5a8b272f489b5fd35538ab1af5fdf57ea6d3aeb7/build.sbt#L93

You can use `-scalajs-mapSourceURI` if Scala 3

fthomas: Thanks @xuwei-k! I'll update it.

I more or less copied the code from typelevel/cats where it is also not set for Scala 3 builds. We should update it there too.

PR #919, opened 2021-03-05
Add inference for #917

kubukoz:
Closes #917.

I will need help with naming 😅 

Also, needs to be copied to 3.0+, I thought to take care of it when this is considered reasonable.

Issue #917, opened 2021-03-05
Combining boolean predicates


matwojcik:
I have some problem when playing with refines using boolean predicates. 

```scala
  object test {
    import eu.timepit.refined.auto._
    type LongOrDouble = NonEmpty And (ValidLong Or ValidDouble)

    val SomeLong: String Refined (NonEmpty And ValidLong) = "11"
    val SomeLongOrDouble: String Refined LongOrDouble = SomeLong


  }
```

which fails with

```
[error] type mismatch;
[error]  found   : eu.timepit.refined.api.Refined[String,eu.timepit.refined.boolean.And[eu.timepit.refined.collection.NonEmpty,eu.timepit.refined.string.ValidLong]]
[error]     (which expands to)  eu.timepit.refined.api.Refined[String,eu.timepit.refined.boolean.And[eu.timepit.refined.boolean.Not[eu.timepit.refined.collection.Empty],eu.timepit.refined.string.ValidLong]]
[error]  required: eu.timepit.refined.api.Refined[String,test.LongOrDouble]
[error]     (which expands to)  eu.timepit.refined.api.Refined[String,eu.timepit.refined.boolean.And[eu.timepit.refined.boolean.Not[eu.timepit.refined.collection.Empty],eu.timepit.refined.boolean.Or[eu.timepit.refined.string.ValidLong,eu.timepit.refined.string.ValidDouble]]]
[error]   val SomeLongOrDouble: String Refined LongOrDouble = SomeLong
[error]                                                       ^

```

```scala
object test2 {
  import eu.timepit.refined.auto._
  type LongOrDouble = (NonEmpty And ValidLong) Or (NonEmpty And ValidDouble)

  val SomeLongOrDouble: String Refined LongOrDouble = "11"

  val SomeNonEmpty: String Refined NonEmpty = SomeLongOrDouble
}

```

which fails with

```
[error] type mismatch;
[error]  found   : eu.timepit.refined.api.Refined[String,test2.LongOrDouble]
[error]     (which expands to)  eu.timepit.refined.api.Refined[String,eu.timepit.refined.boolean.Or[eu.timepit.refined.boolean.And[eu.timepit.refined.boolean.Not[eu.timepit.refined.collection.Empty],eu.timepit.refined.string.ValidLong],eu.timepit.refined.boolean.And[eu.timepit.refined.boolean.Not[eu.timepit.refined.collection.Empty],eu.timepit.refined.string.ValidDouble]]]
[...]

PR #927, opened 2021-03-10
Add inference for #917

matwojcik:
Agreed with @kubukoz that this PR will replace his https://github.com/fthomas/refined/pull/919.

I've suggested some naming, made one of inferences more generic, and added scala 3.0+ version.

### DIFF
--- a/modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/boolean.scala
+++ b/modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/boolean.scala
@@ -277,8 +277,8 @@ private[refined] trait BooleanInference0 extends BooleanInference1 {
     p1.adapt("substitutionInConjunction(%s)")
 
   implicit def disjunctionTautologyElimination[A, B, C](implicit
-      p1: B ==> C,
-      p2: A ==> C
+      p1: A ==> C,
+      p2: B ==> C
   ): (A Or B) ==> C =
     Inference.combine(p1, p2, "disjunctionElimination")
 
--- a/modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/boolean.scala
+++ b/modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/boolean.scala
@@ -273,11 +273,14 @@ private[refined] trait BooleanInference0 extends BooleanInference1 {
   implicit def conjunctionCommutativity[A, B]: (A And B) ==> (B And A) =
     Inference.alwaysValid("conjunctionCommutativity")
 
-  implicit def dontknownhowtocallthis[A, B, C](implicit p1: B ==> C): (A And B) ==> (A And C) =
-    p1.adapt("aaaaaaaaaa")
+  implicit def substitutionInConjunction[A, B, C](implicit p1: B ==> C): (A And B) ==> (A And C) =
+    p1.adapt("substitutionInConjunction(%s)")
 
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/test/scala-3.0-/eu/timepit/refined/BooleanInferenceSpec.scala
```diff
@@ -5,6 +5,8 @@ import eu.timepit.refined.api.Inference
 import eu.timepit.refined.boolean._
```
kubukoz, line 5: I'm kinda worried that this file is only in `scala-3.0-` 🤔

modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/boolean.scala
```diff
@@ -273,6 +273,15 @@ private[refined] trait BooleanInference0 extends BooleanInference1 {
   implicit def conjunctionCommutativity[A, B]: (A And B) ==> (B And A) =
     Inference.alwaysValid("conjunctionCommutativity")
 
+  implicit def substitutionInConjunction[A, B, C](implicit p1: B ==> C): (A And B) ==> (A And C) =
+    p1.adapt("substitutionInConjunction(%s)")
+
+  implicit def disjunctionTautologyElimination[A, B, C](implicit
+      p1: B ==> C,
```
kubukoz, line 280: why this ordering? does `B ==> C` need to go first?

modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/boolean.scala
```diff
@@ -273,6 +273,15 @@ private[refined] trait BooleanInference0 extends BooleanInference1 {
...
+      p2: A ==> C
+  ): (A Or B) ==> C =
+    Inference.combine(p1, p2, "disjunctionElimination")
```
kubukoz, line 283: `Combine` is "and", did you mean to make this use "or"? (as we discussed, probably rqeuires a new method)

matwojcik: Yep, but we were wrong :) In order to imply that `A Or B` can be replaced by `C`, both `A` and `C` needs to be equivalents of `C`.


Here is the truth table for OR option:

```
  A B C  │  ((B ↔ C) ∨ (A ↔ C)) → ((A ∨ B) ↔ C)
  ───────┼─────────────────────────────────────
  1 1 1  │      1    1    1    *1     1    1   
  1 1 0  │      0    0    0    *1     1    0   
  1 0 1  │      0    1    1    *1     1    1   
  1 0 0  │      1    1    0    *0     1    0   
  0 1 1  │      1    1    0    *1     1    1   
  0 1 0  │      0    1    1    *0     1    0   
  0 0 1  │      0    0    0    *1     0    0   
  0 0 0  │      1    1    1    *1     0    1  
```

modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/boolean.scala
matwojcik: nope, will swap ;)

matwojcik:
@fthomas could you review this please?

fthomas (approved):
LGTM. Thanks @matwojcik and @kubukoz!

PR #943, opened 2021-04-12
Add Contentsquare as adopter

Bertrand31:
[Contentsquare](https://contentsquare.com/) uses refined in production.

I recently wrote a piece about how and why we use the library: https://engineering.contentsquare.com/2021/scala-refined-types/


fthomas:
This looks great. Thank you!

PR #971, opened 2021-05-17
Prepare refined-cats sources for Scala 3

fthomas:


qwbarch:
Hey Thomas, this PR brings ``refined-cats`` to Scala 3 right?  
``"eu.timepit" %% "refined-cats" % "0.9.26"`` seems to be missing for Scala 3

fthomas:
No, refined-cats is not yet published for Scala 3. The tests of that module currently do not compile with Scala 3. This PR only changed the non-test sources so that they compile with Scala 3.

PR #986, opened 2021-06-11
Add infix operator to refined and boolean types

### LINKED ISSUE #977: Alphanumeric type is not declared infix
When using refined with Scala 3, the compiler complains about types not being declared as infix. For example:
```scala
val helloWorld: String Refined NonEmpty
```
This will cause the compiler to complain:
```
Alphanumeric type Refined is not declared `infix`; it should not be used as infix operator.
The operation can be rewritten automatically to `Refined` under -deprecation -rewrite.
Or rewrite to prefix syntax Refined[...] manually.bloop
```

qwbarch:
Fixes #977. 
I added ``infix`` for all the types the compiler was complaining about when ``-source:future`` is enabled in the core module.


fthomas (approved):
Thanks @qwbarch!

PR #994, opened 2021-07-09
Move inferences to lower level of implicits, fixes #990

### LINKED ISSUE #990: Inconsistency between 0.9.25 and 0.9.26 versions - possible regression
Hey,
it seems that there is inconsistency between 0.9.25 and 0.9.26 versions of the library
Below you can find code snippet showing that with versions <= 0.9.25 `Concrete` type could be passed to function which expects `Base`
With 0.9.26 it's not possible (compilation error)
Is that intentional?

```
  type BaseRefinement = And[Size[Equal[10]], Forall[LetterOrDigit]]
  type Base = String Refined BaseRefinement

  type ConcreteRefinement = And[StartsWith[W.`"001"`.T], BaseRefinement]
  type Concrete = String Refined ConcreteRefinement

  def funBase(b: Base): Unit = println(b)
  def funConcrete(c: Concrete): Unit = println(c)

  val b: Base = "1234567890"
  val c: Concrete = "0012345678"

  funBase(b)
  funBase(c) // with 0.9.26 doesn't compile, with 0.9.25 works fine

  //    funConcrete(b) // doesn't compile on both versions - that's fine
  funConcrete(c)
```

Full code snippet ready to compile can be found here (https://gist.github.com/lewapek/3cbf003a000bbe9bf479607e9967d201)

Thanks :)

matwojcik:
It fixes #990

### DIFF
--- a/modules/core/shared/src/main/scala-3.0+/eu/timepit/refined/boolean.scala
+++ b/modules/core/shared/src/main/scala-3.0+/eu/timepit/refined/boolean.scala
@@ -245,15 +245,6 @@ private[refined] trait BooleanInference0 extends BooleanInference1 {
   implicit def conjunctionCommutativity[A, B]: (A And B) ==> (B And A) =
     Inference.alwaysValid("conjunctionCommutativity")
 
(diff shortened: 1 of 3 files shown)


fthomas (approved):
Many thanks @matwojcik!

PR #996, opened 2021-07-20
Update Scaladex badge on ReadMe to show Scala version support

rtyley:
This updated badge summarises which versions of Scala are supported by `refined` (and what the latest `refined` version is for each of those Scala versions):

[refined Scala version support](https://index.scala-lang.org/fthomas/refined/refined)

More details on the badge format: https://github.com/scalacenter/scaladex/pull/660


fthomas (approved):
Thanks @rtyley!

PR #1015, opened 2021-09-16
Update scalafmt-core to 3.0.3

sh0hei:


.scalafmt.conf
```diff
@@ -1,6 +1,7 @@
-version = 2.7.5
+version = 3.0.3
+
 maxColumn = 100
-docstrings = JavaDoc
+docstrings.style = keep
```
sh0hei, line 4: It should really be `docstrings.style = Asterisk`. 
cf. https://scalameta.org/scalafmt/docs/configuration.html#docstringsstyle--asterisk
However, since the example code in Scaladoc is formatted, it is set to `docstrings.style = keep`.

スクリーンショット 2021-09-16 22 11 03

fthomas (approved):
That's great. Thank you, @sh0hei!

PR #1058, opened 2022-01-10
Added ValidBoolean predicate to string

dani0854:
Added `ValidBoolean` predicate to string since I didn't find an existing predicate provided by core. Boolean is a common type.
Also renamed function in tests `validNumeric` to `validType` since it work not only for numerics.
Moreover, `toBoolean` infers `Trimmed`, probably the same for numeric predicates.


dani0854:
@fthomas
Should I rebase and resolve conflicts? Will it be considered for merging, or there is no point?

PR #1088, opened 2022-05-04
Add BigDecimal instances for cats

FunFunFine:
This MR add missing Semigroup and Monoid instances for Pos/NegBigDecimal.


FunFunFine:
Hi @fthomas , can you merge this MR please?

fthomas:
@FunFunFine Could you add tests for these instances to https://github.com/fthomas/refined/blob/master/modules/cats/shared/src/test/scala/eu/timepit/refined/cats/CommutativeSemigroupAndMonoidLawTests.scala ?

FunFunFine:
@fthomas it's been a while, but here it goes.

fthomas:
Thanks @FunFunFine!

Can we also add `commutativeSemigroup` / `commutativeMonoid` tests like this:
```scala
checkAll("CommutativeSemigroup[PosBigDecimal]", CommutativeSemigroupTests[PosBigDecimal].commutativeSemigroup)
```
Or do they also fail because those instances are only  approximately associative, too (like `PosDouble` for example)?

FunFunFine:
The compilation fails with the following error when I try to add this test:
```
diverging implicit expansion for type org.scalacheck.Arbitrary[(scala.math.BigDecimal, eu.timepit.refined.numeric.Greater[shapeless._0])]
starting with method arbTuple2 in trait ArbitraryArities
  checkAll("CommutativeSemigroup[PosBigDecimal]", CommutativeSemigroupTests[PosBigDecimal].commutativeSemigroup)
```
I don't really understand what the problem is because it works for other types, but not for BigDecimal apparently.

fthomas:
Oh, that fails because we lack `Arbitrary` instances for refined `BigDecimal` types. I'm fine with merging this without these tests then.

FunFunFine:
Someone who has access to merging needs to merge it

Issue #1092, opened 2022-06-08
Pureconfig not published for scala3


crispywalrus:
Pureconfig publishes a scala 3 artifact but refined-pureconfig isn't published.

fthomas:
I'd like to keep this open until refined-pureconfig is published for Scala 3.

PR #1093, opened 2022-06-08
enable scala3 publishing for refined-pureconfig

crispywalrus:
by moving it from the 2.12/2.13 list to the full list.

addresses #1092

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,7 @@ val moduleCrossPlatformMatrix: Map[String, List[Platform]] = Map(
 )
 
 val moduleCrossScalaVersionsMatrix: (String, Platform) => List[String] = {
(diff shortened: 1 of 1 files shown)


fthomas:
I think we can do this if we stop using pure config-generic for the tests.

crispywalrus:
I thought so to, but pureconfig core pulls it in (thus making pureconfig not scala 3 compatible, seems like that shows how slow uptake has been) I have a pureconfig that's hacked to make a pureconfig-generic as a local published module which is what fooled me into thinking I could get this out

fthomas:
@crispywalrus I don't see in https://repo1.maven.org/maven2/com/github/pureconfig/pureconfig-core_3/0.17.1/pureconfig-core_3-0.17.1.pom that it pulls in pureconfig-generic. How does this happen? Is there a PureConfig issue about this problem?

PR #1098, opened 2022-06-26
Use `scala-xml` 2.x with Scala 2.13+ to fix conflics with sbt-scoverage

guizmaii:
See https://github.com/scoverage/sbt-scoverage/issues/439


guizmaii:
@fthomas Locally, the tests pass on Scala 2.13 with this change

guizmaii:
Seeing https://github.com/fthomas/refined/pull/1099, I think we don't need my PR anymore. Thanks @fthomas

Issue #1090, opened 2022-05-22
Publish for Scala native?


adampauls:
I don't see a Scala native binary on Maven since 0.9.12 (before Scala 3). I see there is still native support in build.sbt; is it broken for some reason, or perhaps it's simply not been published? I'm trying to get native support for circe: https://github.com/circe/circe/pull/1929.

fthomas:
Support for Scala Native has been dropped in https://github.com/fthomas/refined/pull/740 because it was abandoned at that time. I'd merge a PR that brings support for it back to refined.

PR #1103, opened 2022-07-06
Restore some Native builds

armanbilge:
Closes https://github.com/fthomas/refined/issues/1090.

This restores the Native builds for core, scalacheck, and shapeless modules. This is sufficient to unblock Circe Native.

There was a boxing related bug when running the Cats tests, so no Native for that today. Scodec could also get Native in the future (bits is cross-published, but not core yet).


fthomas (approved):
Many thanks, @armanbilge!

armanbilge:
Thanks for releasing! :)

fthomas:
You're welcome!

PR #1152, opened 2023-02-13
Add Dassault Systems to list of adopters

batakpout:
We are using refined in our [Medidata](https://www.medidata.com/en/) Product :)


fthomas (approved):
Thanks @batakpout!

Issue #1156, opened 2023-03-02
refined-scalacheck all values discarded


err0r500:
Hello,
I couldn't find any documentation about how to use refined-scalacheck and I struggle with refinements when they become a bit too specific (looks like they act as a filter discarding all values)

I'm new to scala and I suspect that the implicit resolution is not what it should be but I couldn't figure out how to make it work.

```scala
import org.scalacheck.Properties
import org.scalacheck.Prop.forAll

import eu.timepit.refined.auto.autoUnwrap
import eu.timepit.refined.api.Refined
import eu.timepit.refined.predicates.string.StartsWith
import eu.timepit.refined.scalacheck.string.startsWithArbitrary
import eu.timepit.refined.scalacheck.any.arbitraryFromValidate // <- my main suspect

object StringSpecification extends Properties("String") {
  property("startsWith") = forAll { (str: String Refined StartsWith["a"]) =>
    str.charAt(0).equals("a")
  }
}

// result -> String.startsWith: Gave up after only 0 passed tests. 501 tests were discarded.
```

Any hint about how I'm supposed to use it ?

fthomas:
It works if you remove the `arbitraryFromValidate` import because `startsWithArbitrary` is already imported: https://scastie.scala-lang.org/3zOprdJ5R7CRdcckNGfFVQ

If your refinement becomes more complex, you probably won't have luck with `arbitraryFromValidate` because - as you already said - it is just a filter that discards not matching values.

err0r500:
thanks @fthomas !
I use scala3 and if I just switch the target from your example, it doesn't compile : [https://scastie.scala-lang.org/1J4smLVBQ86ckV9qgvf9pg](https://scastie.scala-lang.org/1J4smLVBQ86ckV9qgvf9pg)

with the error : 
```
No given instance of type org.scalacheck.Arbitrary[
  eu.timepit.refined.api.Refined[String, 
    eu.timepit.refined.string.StartsWith[("a" : String)]
  ]
] was found for parameter a1 of method forAll in object Prop.
I found:

    org.scalacheck.Arbitrary.arbContainer2[eu.timepit.refined.api.Refined, String, 
      eu.timepit.refined.string.StartsWith[("a" : String)]
    ](
      org.scalacheck.Arbitrary.arbContainer2[Tuple2, String, 
        eu.timepit.refined.string.StartsWith[("a" : String)]
      ](
[...]

PR #1157, opened 2023-03-03
Fix/remove shapeless from scalacheck string module

err0r500:
Removes shapeless from `eu.timepit.refined.scalacheck` string arbitraries. (as discussed in #1156 ) in order to bring scala3 compatibility.

**NB :**
- I had to remove `NativePlatform` from scalacheck build matrix in order to get it to work :( 
- I took the liberty to add 2 tests in order to check that changes actually fix the issue


err0r500:
wow, looks like I broke everything !

fthomas:
Yeah, we need to duplicate the sources for Scala 2 and Scala 3 to preserve binary compatibility. My last commit did that.

PR #1162, opened 2023-03-10
Add pureconfig for scala3 [fixes fthomas/refined#1092]

ivan-klass:
### Added `ConfigReader` and `ConfigWriter` for refined types

I've kept 2.13 signature as is in order to keep binary compatibility - I'll be happy if one can help checking if it is actually maintained.

Also, now refined `ConfigReader` and `ConfigWriter` can be obtained independently from `ConfigConvert` (more basic instances have lower priority when resolving). 
I tend to think that `ConfigWriter` is used much more rarely in practice then `ConfigReader`. But with previous code it was impossible to get `ConfigReader[T Refined P]` if you only have `ConfigReader[T]` but no `ConfigWriter[T]`


fthomas:
Thanks for the PR, @ivan-klass!

The Scala 3 build failed with:
```
[error] -- [E008] Not Found Error: /home/runner/work/refined/refined/modules/pureconfig/shared/src/test/scala-3.0+/eu/timepit/refined/pureconfig/SpecDerivedInstances.scala:17:36 
[error] 17 |    ConfigValueFactory.fromMap(JMap.of("value", cfg))
[error]    |                               ^^^^^^^
[error]    |                        value of is not a member of object java.util.Map
```

build.sbt
```diff
@@ -219,9 +219,14 @@ lazy val pureconfig = myCrossProject("pureconfig")
...
-      "com.github.pureconfig" %% "pureconfig-generic" % pureconfigVersion % Test
-    )
+      "com.github.pureconfig" %% "pureconfig-core" % pureconfigVersion
+    ) ++ (CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, 12 | 13)) =>
+        Seq(
+          "com.github.pureconfig" %% "pureconfig-generic" % pureconfigVersion % Test
+        )
+      case _ => Seq.empty
+    })
```
fthomas, line 229: Would it be much work to rewrite the Scala 2 tests without `pureconfig-generic`? It would be nice if we could get rid of `pureconfig-generic` and this complication in the build.

ivan-klass: Sure thing, will do now.


> ```
> [error]    |                        value of is not a member of object java.util.Map
> ```

Fixed. My bad, didin't test with old Java.

As for generic, I'll be happy to fix soon

ivan-klass:
Linking #1092 to this  (I wonder why PR title doesn't work [as described by docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

fthomas:
> I've kept 2.13 signature as is in order to keep binary compatibility - I'll be happy if one can help checking if it is actually maintained.

The CI checks this with MiMa. I just checked this locally again to make sure that it is compatible:
```
sbt:refined> pureconfigJVM/mimaPreviousArtifacts
[info] Set(eu.timepit:refined-pureconfig:0.10.0, eu.timepit:refined-pureconfig:0.10.1, eu.timepit:refined-pureconfig:0.10.2)
sbt:refined> pureconfigJVM/mimaReportBinaryIssues
[info] compiling 45 Scala sources to refined/modules/core/jvm/target/scala-2.13/classes ...
[info] compiling 3 Scala sources to refined/modules/pureconfig/jvm/target/scala-2.13/classes ...
[success] Total time: 7 s, completed Mar 12, 2023, 9:45:50 AM
```
Seems to be fine.

modules/pureconfig/shared/src/main/scala-3.0-/eu/timepit/refined/pureconfig/TypeDescribeVersionSpecific.scala
```diff
@@ -0,0 +1,7 @@
+package eu.timepit.refined.pureconfig
+
+private[pureconfig] trait TypeDescribeVersionSpecific {
+  final type DescribeType[A] = scala.reflect.runtime.universe.WeakTypeTag[A]
```
fthomas, line 4: That's a nice trick to keep bin compat! :+1:

modules/pureconfig/shared/src/main/scala-3.0+/eu/timepit/refined/pureconfig/TypeDescribeVersionSpecific.scala
```diff
@@ -0,0 +1,19 @@
...
+    import quotes.reflect.*
+    '{ new DescribeType[T](${ Expr(TypeRepr.of[T].show) }) }
+  }
+
+  inline given describeType[T]: DescribeType[T] = ${ macroImpl[T] }
+
+private[pureconfig] trait TypeDescribeVersionSpecific {
+
+  protected def getDescription[A](d: DescribeType[A]): String = d.repr
+}
```
fthomas, line 19: This looks similar to [`TypeName` in ciris](https://github.com/vlovgr/ciris/blob/v3.1.0/modules/refined/shared/src/main/scala-3/ciris/refined/internal/TypeNameRuntimePlatform.scala) and the (unmaintained?) https://github.com/tpolecat/typename project.

Is there a difference between `TypeRepr.of[T].show` as used here and `Type.show[A]`? According to the [docs](https://github.com/lampepfl/dotty/blob/3.2.2/library/src/scala/quoted/Type.scala#LL17C5-L17C24)  `Type.show[A]` is the same as `TypeTree.of[T].show`.

ivan-klass: @fthomas The difference is between `Type[T]` and `TypeRepr[T]` - they're different thing, but I'm not sure we care about differences here.
I've found [this article](https://softwaremill.com/scala-3-macros-tips-and-tricks/) extremely helpful, I've used it as a starting point when implementing, there's a note about both `Type` and `TypeRepr`

As for `show`ing - it's all about using `Printer`s - but `Type.show[A]` signature is fixed to the default one, while `TypeRepr[T].show` accepts a given `Printer` (if not in scope, resolves to `Printer.TypeReprCode`). 

So, finally -

 - `Type.show[A]` is `Printer.TreeCode.show(summon[TypeTree[T]])`
 - `TypeRepr.of[T].show` is `Printer.TypeReprCode.show(summon[TypeRepr[T]])`

Other available printers can be found here in [dotty sources](https://github.com/lampepfl/dotty/blob/528d93189480570532ee20b27814dcc16a014775/library/src/scala/quoted/Quotes.scala#L5057-L5086) - let me know if it makes sense to try another one for prettier results

fthomas: Many thanks for the explanation and links, @ivan-klass! There is no need to try another printer here. I just hope that in the future this kind of macro does not need to be reinvented in every library. :crossed_fingers:

fthomas (approved):
Looks great. Thanks again, @ivan-klass!

PR #1248, opened 2024-02-04
remove shapeless imports from initialCommands

xuwei-k:
- https://github.com/fthomas/refined/pull/1193


fthomas (changes requested):
Thanks @xuwei-k! I'd prefer to just remove the shapeless imports from `initialCommands` altogether to make the build simpler.

fthomas (approved):

PR #1274, opened 2024-06-01
Update Scala Native to 0.5, revert to Scala 3.3

keynmol:
I noticed that Scala was updated to 3.4.1 (https://github.com/fthomas/refined/pull/1261) perhaps unintentionally - libraries in general are recommended to stay on LTS I believe, unless they use features only present in Scala Next.


fthomas:
Thanks for the PR, @keynmol! At least a `sbt githubWorkflowGenerate` is required to fix the build.

fthomas:
Released in 0.11.2.

PR #1317, opened 2024-12-19
De-prioritize `intWitnessAs` implicit on Scala 3

armanbilge:
There are already several specialized instances e.g. `intWitnessAsLong`. These should be higher priority than this generic instance that relies on `Numeric`.

Annoyingly, starting in Scala 3.6 the implicit resolution changes cause the following warning. Fixing the prioritization should address the warning.

```
[warn]    |Given search preference for eu.timepit.refined.internal.WitnessAs[(0 : Int), Long] between alternatives
[warn]    |  (eu.timepit.refined.internal.WitnessAs.intWitnessAsLong :
[warn]    |  [A <: Int]: eu.timepit.refined.internal.WitnessAs[A, Long])
[warn]    |and
[warn]    |  (eu.timepit.refined.internal.WitnessAs.intWitnessAs :
[warn]    |  [B, A <: Int]
[warn]    |    (implicit wa: ValueOf[A], ta: eu.timepit.refined.internal.ToInt[A], nb:
[warn]    |      Numeric[B]): eu.timepit.refined.internal.WitnessAs[A, B]
[warn]    |)
[warn]    |will change.
[warn]    |Current choice           : the first alternative
[warn]    |New choice from Scala 3.7: the second alternative
```

---

Alternatively, perhaps this implicit was ported in error. The equivalent on the Scala 2 side is for `Nat`, which isn't available for Scala 3, so perhaps was swapped for `Int`?

https://github.com/fthomas/refined/blob/20df7ccb625448023a7ecddf9af84ad4f489415f/modules/core/shared/src/main/scala-3.0-/eu/timepit/refined/internal/WitnessAs.scala#L31-L36

### DIFF
--- a/modules/core/shared/src/main/scala-3.0+/eu/timepit/refined/internal/WitnessAs.scala
+++ b/modules/core/shared/src/main/scala-3.0+/eu/timepit/refined/internal/WitnessAs.scala
@@ -25,14 +25,7 @@ import scala.compiletime.{constValue, error}
 final case class WitnessAs[A, B](fst: A, snd: B)
 
 object WitnessAs extends WitnessAs1 {
-  def apply[A, B](implicit ev: WitnessAs[A, B]): WitnessAs[A, B] = ev
-
-  implicit def intWitnessAs[B, A <: Int](implicit
-      wa: ValueOf[A],
-      ta: ToInt[A],
-      nb: Numeric[B]
-  ): WitnessAs[A, B] =
-    WitnessAs(wa.value, nb.fromInt(ta.apply()))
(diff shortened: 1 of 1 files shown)


fthomas:
> Alternatively, perhaps this implicit was ported in error. The equivalent on the Scala 2 side is for Nat, which isn't available for Scala 3, so perhaps was swapped for Int?

This instance was once for `Nat` on Scala 3:

https://github.com/fthomas/refined/commit/00dcf50b621524b1ab8f08f90fe1a12dec7a0064#diff-5b011ae4abb8c08722548566381dedd082136bc9353874ece9567afc4a681423L31-R31

If there is no test that fails without it, I'm fine with making not implicit and deprecating it.

PR #1318, opened 2024-12-19
optimize `RefinedTypeOps.unapply`

xuwei-k:
avoid Intermediate `Either` objects


modules/core/shared/src/main/scala-3.0+/eu/timepit/refined/api/RefinedTypeOps.scala
```diff
@@ -25,8 +25,11 @@ class RefinedTypeOps[FTP, T](implicit rt: RefinedType.AuxT[FTP, T]) extends Seri
   def from(t: T): Either[String, FTP] =
     rt.refine(t)
 
-  def unapply(t: T): Option[FTP] =
-    from(t).toOption
+  def unapply(t: T): Option[FTP] = {
+    val res = rt.validate.validate(t)
+    if (res.isPassed) Option(rt.refType.unsafeWrap(t).asInstanceOf[FTP])
```
fthomas, line 30: Is the null check via `Option.apply` intentional? The previous implementation did not do that.

PR #1378, opened 2025-09-20
enable doctest for scala-js and scala-native

xuwei-k:
- doctest support scala-js and scala-native in latest version https://github.com/sbt-doctest/sbt-doctest/issues/52


fthomas:
Thanks!

